PR #84, opened 2026-03-25
feat(events): add #[event] macro and IDL event support

bristinWild:
Adds #[event] macro and IDL event schema support, enabling spel tx decode to do full typed event decoding from transaction receipts.
Changes:

IdlEvent struct added to SpelIdl (name, discriminant, fields: Vec<IdlArg>)
events: Vec<IdlEvent> field in SpelIdl — backward-compatible, skipped if empty
#[event(discriminant = N)] proc macro to annotate event payload structs
SpelOutput::with_events() builder that auto-drains the lez-events buffer
generate_idl_json and generate_idl_fn pick up #[event] structs alongside #[instruction] functions
All 90 existing tests passing

Usage:
rust#[event(discriminant = 1)]
pub struct InsufficientFunds {
    pub requested: u128,
    pub available: u128,
}
Relates to: logos-blockchain/logos-execution-zone LP-0012 event system


bristinWild:
Rebased onto upstream/main (d24dbaa, v0.4.0). Conflicts resolved , removed the lez-events crate dependency, keeping IdlEvent and #[event] as self-contained additions to spel-framework-core and spel-framework-macros. All 225 unit tests passing, 0 failures.

vpavlin (changes requested):
Thanks @bristinWild — the schema design here is clean (additive `events: Vec<IdlEvent>` with `skip_serializing_if`, no-op `#[event]` marker consistent with `#[account_type]`). But as it stands the feature is **scaffolding only and doesn't function end-to-end**:

- Nothing on the branch ever **collects** `#[event]`-annotated structs into `SpelIdl.events`. Both constructors (`SpelIdl::new`, `generate_idl_inner`) and the test literals set `events: vec![]`, and there's no scan in `generate_idl!` / `#[lez_program]` / the account-type collector. So `#[event]` is a no-op and the IDL's `events` array is **always empty**.
- The test changes are just mechanical `events: vec![]` additions to satisfy the new field — none assert that a program with `#[event]` structs actually produces populated events.

To land this I'd want:
1. Collection logic that scans `#[event]` structs (discriminant + fields) and populates `SpelIdl.events`, wired into the same paths that collect `#[account_type]` (`spel-framework-core/src/account_types.rs` / `idl_gen.rs` / the `lez_program` expansion).
2. A regression test: a source with an `#[event(discriminant = N)]` struct → assert the generated IDL contains the event with correct name/discriminant/fields.

Happy to help wire the collection into the account-type scanner if useful. Branch is updated onto the LEZ v0.2.0 `main`; holding merge until the collection + test are in.

PR #89, opened 2026-03-30
Make SPEL work with latest LEZ `ProgramOutpu` APIs

0x-r4bbit:


vpavlin:
This is stale and should be already implemented by #90

PR #126, opened 2026-04-09
feat: SpelOutput::execute() with auto-claim support

### LINKED ISSUE #124: Feature: Unified SpelOutput::execute() with auto-claim
## Summary

Design a new unified `SpelOutput::execute()` method that auto-generates `AccountPostState` claims based on `#[account(...)]` attributes, simplifying developer experience.

## Motivation

Currently, developers must manually construct `AccountPostState` with correct `Claim` types:

```rust
// Current painful API:
Ok(SpelOutput::states_only(vec![
    AccountPostState::new_claimed(state.account.clone(), Claim::Pda(seed)),
]))
```

This is error-prone because:
1. Developer must know `Claim::Pda` vs `Claim::Authorized`
2. Developer must compute seed bytes correctly
3. Developer might forget to claim init accounts

## Proposed Solution

New unified method with auto-claim:

```rust
SpelOutput::execute(accounts: Vec<impl IntoPostState>, calls: Vec<ChainedCall>) -> SpelResult
```

**Auto-claim rules:**

| Account attribute | Generated post-state |
|-------------------|---------------------|
| `#[account(init, pda = literal("seed"))]` | `new_claimed(..., Claim::Pda(seed))` |
| `#[account(init, signer)]` | `new_claimed(..., Claim::Authorized)` |
| `#[account(mut)]` | `new()` (update only) |
| `#[account]` (read-only) | `new()` (validate ownership) |

**Developer experience:**

```rust
Ok(SpelOutput::execute(vec![state, owner], vec![]))
```

The macro parses `#[account(...)]` attributes at compile time. At runtime, `execute()` uses this metadata to construct correct `AccountPostState` automatically.

## Migration Path

1. Add `SpelOutput::execute()` as new unified method
2. Mark `states_only()` and `with_chained_calls()` deprecated
3. In next major release, remove deprecated methods

## Files to Modify

1. `spel-framework-core/src/spel_output.rs` — new `execute()` method
2. `spel-framework-macros/src/lib.rs` — parse attributes, emit metadata
3. `spel-framework-core/src/types.rs` — `IntoPostState` trait

## References

- LEZ v0.2.0-rc1 claim validation rules
- Issue #123 (related: Do we need explicit init?)
- PR #126 (implementation)

jimmy-claw:
## Summary

New unified SpelOutput::execute() method that auto-generates AccountPostState claims based on #[account(...)] attributes.

## Changes

1. NEW: spel-framework-core/src/spel_output.rs
   - AutoClaim enum with variants: Claimed(Claim::Pda), Claimed(Claim::Authorized), None
   - SpelOutput::execute(accounts, calls) - generic method that auto-claims
   - 8 unit tests passing

2. spel-framework-core/src/types.rs
   - Added IntoPostState trait
   - Deprecated states_only() and with_chained_calls()

3. spel-framework-macros/src/lib.rs
   - Generates __claims_{fn_name}() per instruction with auto-claim metadata

## Auto-Claim Rules

| Attribute | Generated |
|-----------|-----------|
| #[account(init, pda = ...)] | Claim::Pda(seed) |
| #[account(init, signer)] | Claim::Authorized |
| #[account(mut)] | new() |
| #[account] (read-only) | new() |

Closes #124


marclawclaw:
## Summary

This PR adds `SpelOutput::execute()` with auto-claim support, replacing the manual `states_only()` / `with_chained_calls()` API. The macro now generates `__claims_*()` functions per instruction that derive claim metadata from `#[account(...)]` constraints, and an `ExecuteTransformer` rewrites `SpelOutput::execute(...)` calls to `execute_with_claims(...)` at compile time. Also adds PDA validation in `__validate_*()`, a `ToSeed` trait for typed seed conversion, and filtering of non-owned accounts from program output.

Well-structured PR with good test coverage of the new PDA paths. A few issues below.

## Correctness

**Bug — `PdaSeedDef::Account` handled incorrectly in claim generation** (`spel-framework-macros/src/lib.rs` ~L897):

In `generate_single_claim_expr`, the `Account` variant calls `seed_from_str(v)` where `v` is the *account name string* (e.g., `"owner"`). But in the validation function (`generate_validation`, ~L1090), the same `Account` variant correctly resolves to `accounts[idx].account_id.value()` — the actual runtime account ID. This means the PDA seed in the claim will be `SHA256("owner"...)` while validation checks against the real account ID. These will never match for `account(...)` PDA seeds.

```rust
// In generate_single_claim_expr — WRONG:
PdaSeedDef::Account(v) => {
    let val = v.clone();
    quote! { &spel_framework::pda::seed_from_str(#val) }  // uses string name
}

// In generate_validation — CORRECT:
PdaSeedDef::Account(path) => {
    let account_idx = ...;
    quote! { let #var = *accounts[#account_idx].account_id.value(); }  // uses real ID
}
```

The claim generation should look up the account by index and use `accounts[idx].account.clone()` (or the account ID) at runtime, similar to validation. No test covers `account(...)` PDA seeds in claims, so this goes undetected.

**Double evaluation of accounts expression** (`ExecuteTransformer`, dynamic/rest branch ~L791):

When the instruction has rest accounts and the accounts arg is an arbitrary expression, the transformed code evaluates `#accounts_expr` twice — once as the accounts slice and once for `.len()`. If the expression has side effects (unlikely but possible), this would be surprising. Consider binding to a local variable.

**Filtering logic correctness** (macro `main()` ~L285): The filter `!is_default_owner || pre_is_default || has_claim` looks correct for the stated rule 7 scenario. The logic preserves: (a) program-owned accounts, (b) default-state accounts (init targets), and (c) accounts with claims. Well-documented with comments.

## Code Quality

- Clean separation: `AutoClaim` enum, `IntoPostState` trait, and `SpelOutput::execute` are well-factored.
- The `ExecuteTransformer` using `syn::visit_mut::VisitMut` is a solid pattern for AST rewriting in proc macros.
- `#[deprecated]` on old API methods is a nice migration path.
- `_instruction_words` param in validation functions — unused today (underscore-prefixed). If it's for future use, a brief comment would help.
- Making `main()` and the module `pub` — is this intentional? If it's for testing, worth a comment explaining why.

## Tests

Good coverage overall:
- `ToSeed` trait: identity, u64, u32, String, &str, consistency with `seed_from_str` ✓
- `compute_pda_multi`: matches `compute_pda`, mixed types, single value, three seeds ✓
- `AutoClaim`: `is_claimed`, `pda_from_seeds` single/multi ✓
- `execute` / `execute_with_claims`: auto-claims, zip correctness, empty, mismatch panic, raw post-states ✓
- PDA validation: correct/incorrect PDAs for single-arg, multi-seed, u64+u32 args, String args ✓
- IDL: updated counts for new instructions, seed metadata assertions ✓

**Missing**:
- No test for `PdaSeedDef::Account(...)` seeds (the `account("other_account")` PDA seed variant) — this is where the claim bug lurks.
- No test for the output filtering logic in `main()` (the non-owned account filter). This is generated code so harder to unit-test, but an integration test would catch regressions.
- No test for the `ExecuteTransformer` with the "arbitrary expression" fallback path (non-`vec![...]` fixed-account case).

## API Design

- `SpelOutput::execute()` accepting `impl IntoIterator<Item = impl IntoPostState>` is flexible — supports `Vec<(Account, AutoClaim)>`, `Vec<AccountPostState>`, etc.
- `execute_with_claims()` as the macro-targeted method with a clear panic on length mismatch is appropriate for generated code.
- `AutoClaim::pda_from_seeds(&[&[u8]])` accepts arbitrary-length byte slices but `compute_pda` requires exactly `&[u8; 32]`. The hash will match when seeds are 32 bytes, but the type signature doesn't enforce this. Consider `&[&[u8; 32]]` for consistency, or document the 32-byte requirement.
- The `ToSeed` trait is a clean abstraction. Missing `u8`, `u16`, `i32`, `i64` impls — probably fine for now but worth noting as a future extension point.

## Suggestions

**Critical**: Fix `PdaSeedDef::Account` handling in `generate_single_claim_expr` to use the actual account ID at runtime (matching what validation does), and add a test for it.

**Minor**: Consider binding `accounts_expr` to a `let` in the `ExecuteTransformer` rest branch to avoid double evaluation.

PR #129, opened 2026-04-11
fix(cli): parse PDA seed args through IDL type system

fryorcraken:
## Summary

- **fix(cli): parse PDA seed args through IDL type system** — `compute_pda_command` stored all seed args as `ParsedValue::Str(raw)` instead of routing through `parse_value()` with the IDL type. This meant `[u8; 32]` args were never parsed as `ByteArray`, producing wrong PDA hashes and "type mismatch in risc0 serialization" warnings.
- **fix(macros): emit structured IdlType variants** — `generate_idl_fn` wrapped every arg type in `IdlType::Primitive(string)`, so `[u8; 32]` became `Primitive("[u8; 32]")` instead of `Array{Primitive("u8"), 32}`. This caused `parse_value` to miss the Array dispatch, producing Raw values that broke PDA seed computation.
- **test(cli): add risc0 serde roundtrip tests** — Verified that spel's serialization matches the risc0 serde contract used by `nssa_core::program::read_nssa_inputs` at LEZ v0.2.0-rc1. Key finding: risc0 serializes each `u8` as its own `u32` word (zero-extended), not packed 4-per-word. The roundtrip test deserializes spel's output using `risc0_zkvm::serde::Deserializer` (the same code the guest runs) to prove on-chain compatibility.

## Test plan

- [x] Run `cargo +nightly build -p spel` — confirms compilation
- [x] Run `cargo test -p spel-framework-core` — existing PDA tests pass
- [x] Run `cargo +nightly test -p spel` — all 31 tests pass including 5 new serialization tests
- [x] Roundtrip test: spel serialize → risc0 Deserializer → assert values match (proves on-chain compat)
- [x] Manual: `spel pda <account> --<byte-array-arg> <64-char-hex>` produces correct PDA (no serialization warnings)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #130, opened 2026-04-12
refactor(cli): replace hand-rolled serializer with risc0 serde

fryorcraken:
## Summary

- **Fix IDL type parsing**: macro now emits structured `IdlType` variants (`Array`, `Vec`, `Option`) instead of `Primitive("[u8; 32]")` for compound types
- **Fix PDA seed parsing**: CLI now routes PDA seed args through the IDL type system instead of hardcoding `[u8; 32]`
- **Replace hand-rolled serializer with serde**: `serialize_to_risc0` now delegates to `risc0_zkvm::serde::to_vec` via a `DynamicValue` enum with `serde::Serialize` impl, eliminating format divergence by construction
- **Proper error handling**: type mismatches now return `Result<Vec<u32>, SerializeError>` instead of silent `eprintln!` + continue

### Why

The hand-rolled serialization had to exactly match risc0's wire format by coincidence, with no compile-time guarantee. This caused multiple bugs (PR #129, silent truncation on type mismatches). By implementing `serde::Serialize` and delegating to the canonical `risc0_zkvm::serde::to_vec`, format correctness is guaranteed by construction.

### Design

- `DynamicValue` enum mirrors risc0 wire types (Bool, U8, U32, U64, U128, Str, Tuple, Seq, None, Some)
- `to_dynamic_value()` converts `(IdlType, ParsedValue)` pairs, handling all existing type combinations including `Raw(csv)` fallback
- `InstructionData` wrapper uses `serialize_tuple_variant` to write variant_index + fields (risc0 ignores string params, so `""` avoids `&'static str` lifetime issues)

### Key insight

`risc0_zkvm::serde::to_vec(&impl Serialize)` is the canonical serialization path (used by `nssa::program::Program::serialize_instruction`). If we implement `serde::Serialize` for a dynamic value type that mirrors our `ParsedValue`, we can call `to_vec` directly and get format correctness for free. This pattern is well-established (cf. `serde_json::Value`).

### risc0 serde wire format reference (v3.0.5)

| Rust type | serde method | Wire format (u32 words) |
|-----------|-------------|------------------------|
| `u8` | `serialize_u8` | 1 word (zero-extended) |
| `u32` | `serialize_u32` | 1 word |
| `u64` | `serialize_u64` | 2 words (lo, hi) |
| `u128` | `serialize_u128` | 4 words (LE bytes packed) |
| `bool` | `serialize_bool` | 1 word (0 or 1) |
| `String` | `serialize_str` | length + padded bytes |
| `[T; N]` | `serialize_tuple(N)` | N elements, no length prefix |
| `Vec<T>` | `serialize_seq(len)` | length prefix + elements |
| `Option::None` | `serialize_none` | 1 word (0) |
| `Option::Some(v)` | `serialize_some` | 1 word (1) + value |
| enum struct variant | `serialize_struct_variant` | variant_index + fields |

Critical: `[u8; 32]` serializes as 32 individual u32 words (one per byte), NOT packed 4-per-word.

### Out of scope

- **`Defined` type support**: Custom struct/enum args from the IDL. Returns `SerializeError::UnsupportedType` for now.
- **Per-program CLI code generation**: Maximum type safety but requires architecture change.
- **`IdlType::Primitive` as an enum**: Would catch invalid type names at parse time.

## Test plan

- [x] All 5 original serialize roundtrip tests pass (including `serialize_deserialize_roundtrip_with_bytes32` — the CLI-to-guest contract test)
- [x] 12 new roundtrip tests: `[u8; 32]`, all primitives, `Option<u32>` (Some + None), `Vec<u8>`, `Vec<u32>`, `Vec<[u8; 4]>`, error cases
- [x] `risc0_reference_bytes32_format` golden test passes
- [x] Full spel crate: 43 tests pass
- [x] Framework crates: all tests pass (no regressions)
- [x] Manual smoke test (dry-run): submit a transaction with `[u8; 32]` args via CLI


fryorcraken:
I suggest to not merge until proper dogfooding. Will aim to do so ASAP.

fryorcraken:
My dogfooding says go.

PR #131, opened 2026-04-12
feat(spel): enhance --dry-run with full tx summary and JSON output

marclawclaw:
## Summary

Enhances `spel --dry-run` to walk the full resolution path — program load, PDA computation, best-effort nonce fetch — before exiting. Developers can see exactly what would be submitted, in either human-readable text or JSON for scripting.

## Usage

```bash
# Text summary (default)
spel --idl my-idl.json -p target/program.bin \
     submit-transfer --target Owner/abc123... --amount 100 \
     --dry-run

# JSON output — stdout is clean JSON, safe to pipe
spel --idl my-idl.json -p target/program.bin \
     submit-transfer --target Owner/abc123... --amount 100 \
     --dry-run=json > dry-run.json
```

`--dry-run=<fmt>` uses the equals syntax. Unknown formats are rejected.

## Files

- `spel-cli/src/lib.rs` — parse `--dry-run` / `--dry-run=<fmt>`
- `spel-cli/src/cli.rs` — help text
- `spel-cli/src/tx.rs` — restructure `execute_instruction`: program + PDA resolution moved before the dry-run gate; add `print_dry_run_text` / `print_dry_run_json` helpers

## Review notes

- The privacy-preserving submission path requires the program binary (unchanged). If dry-running with `--program <hex>`, only the public-path metadata is visible — privacy-preserving wouldn't have been submittable anyway.
- Nonce fetch is best-effort in dry-run: if the wallet env isn't set, nonces show `(unknown)` / `null` rather than failing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,12 +236,18 @@ spel inspect <account-id> --idl my_program-idl.json --type VaultState --data <bo
 # Show available commands
 spel --idl program-idl.json --help
 
-# Dry run an instruction
-spel --idl program-idl.json --dry-run -p program.bin \
+# Dry run an instruction — resolve everything (PDAs, accounts, serialized data,
+# signer nonces) and print without submitting. Accepts --dry-run (text default),
+# --dry-run=text, or --dry-run=json.
+spel --idl program-idl.json --dry-run -p program.bin -- \
   create-vault --token-name "MYTKN" --initial-supply 1000000
 
+# Machine-readable dry run for scripting / golden tests
+spel --idl program-idl.json --dry-run=json -p program.bin -- \
+  create-vault --token-name "MYTKN" --initial-supply 1000000 | jq .
+
 # Submit a transaction
-spel --idl program-idl.json -p program.bin \
+spel --idl program-idl.json -p program.bin -- \
   create-vault --token-name "MYTKN" --initial-supply 1000000
 
 # Use --program-id instead of binary (skips loading the file)
--- a/spel-cli/src/cli.rs
+++ b/spel-cli/src/cli.rs
@@ -16,7 +16,7 @@ pub fn print_help(idl: &SpelIdl, binary_name: &str) {
     println!("  -p, --program <NAME|HEX|FILE>");
     println!("                             Program name from spel.toml, 64-char hex program ID,");
     println!("                             or path to program binary (or set in spel.toml)");
-    println!("  --dry-run                  Print parsed/serialized data without submitting");
+    println!("  --dry-run[=text|json]      Resolve & print transaction without submitting (text default)");
     println!("  --bin-<NAME> <FILE>        Additional program binary (auto-fills --<NAME>-program-id)");
     println!();
     println!("COMMANDS:");
--- a/spel-cli/src/lib.rs
+++ b/spel-cli/src/lib.rs
@@ -45,7 +45,8 @@ pub async fn run() {
 
     let mut idl_path = String::new();
     let mut program_ref: Option<String> = None; // raw --program value
-    let mut dry_run = false;
+    // None = submit, Some("text") = dry-run text summary, Some("json") = dry-run JSON
+    let mut dry_run: Option<String> = None;
     let mut type_name: Option<String> = None;
     let mut data_hex: Option<String> = None;
     let mut extra_bins: HashMap<String, String> = HashMap::new();
@@ -82,7 +83,15 @@ pub async fn run() {
                 i += 1;
                 if i < args.len() { data_hex = Some(args[i].clone()); }
             }
-            "--dry-run" => { dry_run = true; }
+            "--dry-run" => { dry_run = Some("text".to_string()); }
+            s if s.starts_with("--dry-run=") => {
+                let fmt = &s["--dry-run=".len()..];
+                if fmt != "text" && fmt != "json" {
+                    eprintln!("❌ --dry-run=<fmt>: expected 'text' or 'json', got '{}'", fmt);
+                    process::exit(1);
+                }
+                dry_run = Some(fmt.to_string());
+            }
             s if s.starts_with("--bin-") => {
                 let name = s.strip_prefix("--bin-").unwrap().to_string();
                 i += 1;
@@ -372,7 +381,7 @@ pub async fn run() {
                     }
                     let cli_args = parse_instruction_args(&remaining_args[2..], ix);
                     execute_instruction(
-                        &idl, ix, &cli_args, program_path.as_deref(), program_id_hex.as_deref(), dry_run, &extra_bins,
+                        &idl, ix, &cli_args, program_path.as_deref(), program_id_hex.as_deref(), dry_run.as_deref(), &extra_bins,
                     ).await;
                 }
                 None => {
--- a/spel-cli/src/lib.rs
+++ b/spel-cli/src/lib.rs
@@ -45,8 +45,7 @@ pub async fn run() {
 
     let mut idl_path = String::new();
     let mut program_ref: Option<String> = None; // raw --program value
-    // None = submit, Some("text") = dry-run text summary, Some("json") = dry-run JSON
-    let mut dry_run: Option<String> = None;
+    let mut dry_run: Option<tx::DryRunFormat> = None;
     let mut type_name: Option<String> = None;
     let mut data_hex: Option<String> = None;
     let mut extra_bins: HashMap<String, String> = HashMap::new();
@@ -83,14 +82,16 @@ pub async fn run() {
                 i += 1;
                 if i < args.len() { data_hex = Some(args[i].clone()); }
             }
-            "--dry-run" => { dry_run = Some("text".to_string()); }
+            "--dry-run" => { dry_run = Some(tx::DryRunFormat::Text); }
             s if s.starts_with("--dry-run=") => {
-                let fmt = &s["--dry-run=".len()..];
-                if fmt != "text" && fmt != "json" {
-                    eprintln!("❌ --dry-run=<fmt>: expected 'text' or 'json', got '{}'", fmt);
-                    process::exit(1);
-                }
-                dry_run = Some(fmt.to_string());
+                dry_run = Some(match &s["--dry-run=".len()..] {
+                    "text" => tx::DryRunFormat::Text,
+                    "json" => tx::DryRunFormat::Json,
+                    other => {
+                        eprintln!("❌ --dry-run=<fmt>: expected 'text' or 'json', got '{}'", other);
+                        process::exit(1);
+                    }
+                });
             }
             s if s.starts_with("--bin-") => {
                 let name = s.strip_prefix("--bin-").unwrap().to_string();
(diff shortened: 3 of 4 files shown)


spel-cli/src/lib.rs
```diff
@@ -45,6 +45,7 @@ pub async fn run() {
     let mut program_path = "program.bin".to_string();
     let mut program_id_hex: Option<String> = None;
     let mut dry_run = false;
+    let mut dry_run_output: Option<String> = None;
```
fryorcraken: it's an option, but `json` is the only possible value? or is it a different behaviour if `json` is not passed?

spel-cli/src/tx.rs
```diff
@@ -262,15 +213,215 @@ pub async fn execute_instruction(
         }
     }
 
+    // Check if any account has a Private/ prefix (needed for dry-run privacy warning too)
```
fryorcraken: what is the dry-run privacy warning

fryorcraken: Not fixed.

spel-cli/src/cli.rs
```diff
@@ -13,7 +13,8 @@ pub fn print_help(idl: &SpelIdl, binary_name: &str) {
     println!("OPTIONS:");
     println!("  -i, --idl <FILE>           IDL JSON file");
     println!("  -p, --program <FILE>       Program binary");
-    println!("  --dry-run                  Print parsed/serialized data without submitting");
+    println!("  --dry-run                  Parse and validate locally; show full transaction summary without submitting");
+    println!("  --dry-run-output           With --dry-run, output full transaction as JSON");
```
fryorcraken: is the point to format the output as json? meaning that `--dry-run` is not json? is there a better way to do that?

if that's correct then I recommend

```
--dry-run
--dry-run <text|json> defaults to text
```

README.md
```diff
@@ -245,6 +245,31 @@ spel --idl program-idl.json -p treasury.bin --bin-token token.bin \
...
 ```
 
+## Dry-run: Validate Before Submitting
+
+Before committing to a 15-second block cycle, run any command with `--dry-run` to see the complete local transaction picture:
+
+```sh
+spel --idl my-program-idl.json -p target/program.bin \
+     submit-transfer --target Owner/abc123... --amount 100 \
+     --dry-run
```
fryorcraken: I think it's risky to put the --dry-run at the end, after the guest program and its arguments.
The fact taht spel and guest program argument CAN be mixed seems like a recipe for problems.

marclawclaw (commented):
## Code Review — `feat(spel): enhance --dry-run with full tx summary and JSON output`

**Verdict: REQUEST_CHANGES** (posted as COMMENT — GitHub won't let you request changes on your own PR)

Good feature direction — showing the full tx picture before submission is clearly valuable. Several issues need addressing before merge.

---

### 🔴 Correctness (3 bugs)

**1. `--dry-run json` (space syntax) is dead code**
`lib.rs` has `s if s.starts_with("--dry-run ")` but this arm **can never match**. The shell splits `--dry-run json` into two args `["--dry-run", "json"]`. The `"--dry-run"` exact match fires first, setting `dry_run_format = Some("text")`. Then `"json"` falls through to the subcommand/instruction parsing, where it'll error or be misinterpreted.

Only `--dry-run=json` (equals syntax) actually works. Either:
- Remove the dead `starts_with("--dry-run ")` arm, or
- Peek ahead at `args[i+1]` from the `"--dry-run"` match arm when the next arg is `"json"` or `"text"`

**2. JSON + text both written to stdout — breaks piping**
When `--dry-run=json`, both the JSON blob and the human-readable summary are printed to stdout. The stated use case (`> dry-run.json`) will produce unparseable output. Fix: either suppress the text summary when format is `"json"`, or print the text to stderr.

**3. Rest accounts missing from JSON output**
In the `acc.rest` branch (~line 330 of the diff), entries are pushed to `accounts_summary` and `raw_account_ids` but **not** to `accounts_json`. JSON consumers get an incomplete account list for instructions using variadic rest accounts.

---

### 🟡 Correctness (non-blocking)

**4. Nonce truncation `n.0 as u64`**
`Nonce` is `u128` but is cast to `u64` for display and JSON. If a nonce exceeds `u64::MAX`, this silently wraps. Unlikely in practice today, but worth a `// SAFETY:` comment or an explicit check.

**5. Program binary read twice on submission path**
Step 4 reads and parses the binary into `_program_obj` (unused — note the `_` prefix). After the dry-run exit point, it's re-read from disk. This is functional but wasteful. Consider keeping the first `Program` and reusing it.

**6. `WalletCore::from_env()` called twice on submission path**
Once for the nonce fetch (Step 7), once for tx submission. Could initialize once and reuse.

---

### 🟡 Readability

- `execute_instruction` is now ~450 lines. The step comments help, but consider extracting `print_text_summary()` and `build_json_summary()` as helpers in a follow-up.
- The `has_private` check was moved up but the comment `// has_private already checked above (before Step 8)` could just be removed — the code is self-evident.

---

### 🟢 Architecture

The approach of walking the full resolution path before dry-run exit is sound. Moving program loading, PDA resolution, and nonce fetching before the dry-run return point is the right structural change.

### 🟢 Security

No new attack surface. Nonce fetch gracefully degrades to `(unknown)` in dry-run on failure — appropriate.

### 🟢 Performance

No concerns beyond the double-read noted above (non-blocking).

---

### Reviewer comment status

@fryorcraken's prior feedback:
- ✅ `--dry-run-output` collapsed into `--dry-run [text|json]` as suggested
- ✅ Privacy warning removed (misleading for Solana/NSSA)
- ⚠️ `--dry-run` at end of command line (after guest args) — still present in README examples. fryorcraken flagged this as risky. The latest commit (`64f5773`) changed to `--dry-run=json` which helps for the JSON case, but the plain `--dry-run` example still shows it trailing. Consider moving `--dry-run` before the subcommand in examples.

---

### Summary

Fix the three correctness bugs (dead code arm, stdout mixing, missing rest accounts in JSON), then this is good to merge. The other items are non-blocking suggestions.

marclawclaw:
Pausing this PR — we want to fix the argument parsing in spel first before proceeding with the dry-run enhancements. The 3 blocking issues (space syntax never matches, broken JSON piping, missing rest accounts in JSON) need to be addressed in the context of a broader args parsing fix.

marclawclaw (commented):
## Code Review: feat(spel): enhance --dry-run with full tx summary and JSON output

**Verdict: REQUEST_CHANGES** (posted as comment — can't request changes on own PR)

---

### 1. Correctness

**Nonce truncation — data loss bug** (`tx.rs`, nonce display and JSON):
```rust
Some(Some(n)) => format!("nonce={}", n.0 as u64),
// ...
"nonce": nonces.get(i).and_then(|n| n.map(|x| x.0 as u64)),
```
`Nonce` wraps a `u128`. Casting to `u64` silently truncates high bits. If nonces ever exceed `u64::MAX`, the dry-run output will show wrong values and the JSON will be incorrect. Either display the full `u128`, or add an explicit overflow check/comment explaining why truncation is safe for this protocol.

**`--dry-run` space syntax arm is dead code** (`lib.rs`):
```rust
s if s.starts_with("--dry-run ") => { ... }
```
When the shell tokenizes `--dry-run json`, `args[i]` will be `"--dry-run"` and `args[i+1]` will be `"json"`. The combined string `"--dry-run json"` will never appear as a single element. This arm is unreachable. The `--dry-run` (no value) arm fires instead, defaulting to `"text"`, which means `spel ... --dry-run json` silently ignores `json`. Reviewer fryorcraken flagged this — it's still not fixed.

**Program binary read twice on submission path** (`tx.rs`):
The binary is loaded in Step 4 (into `_program_obj`), but the underscore prefix discards it. Then after the dry-run gate, it's re-read from disk:
```rust
let program_obj = if program_id_hex.is_some() { None } else {
    let program_bytecode = fs::read(program_path)...
```
This is wasteful and introduces a TOCTOU window (binary could change between reads). Use the already-loaded `program_obj` from Step 4 instead of re-reading.

**Transaction summary always prints, even on non-dry-run path:**
The summary block (Step 8/9) runs unconditionally — on submission too. This means production submissions print the full box-drawing summary AND then proceed to submit. This seems unintentional — the old code had a clear `if dry_run { return; }` guard that printed minimal info for non-dry-run. Now the summary is always shown before the `dry_run_format.is_some()` early return check.

**`raw_account_ids` mixes encoding formats** (flagged in PR body):
PDAs use base58 (`format!("{}", id)`) while direct accounts use hex (`format!("0x{}", hex_encode(bytes))`). The JSON schema should be consistent — pick one encoding or add a `format` field. This is acknowledged in the PR body but should be tracked as a follow-up issue.

### 2. Readability

**Massive rustfmt noise**: ~1200 of the 1871 added lines are pure formatting changes across files unrelated to the feature (codegen.rs, ffi_codegen.rs, init.rs, parse.rs, pda.rs, serialize.rs, etc.). This makes the actual feature changes hard to review. Consider splitting the rustfmt pass into a separate commit or PR.

**16 commits for one feature**: The commit history tells the story of iterative debugging (compile fixes, type mismatches, format string fixes) rather than clean logical units. The history was rebased but retains the debugging commits. A squash or interactive rebase to ~3 commits (feat, fix, style) would make the history more useful.

### 3. Architecture

**Nonce fetch in dry-run mode is a side effect**: Dry-run fetches nonces from the sequencer via `wallet_core.get_accounts_nonces()`. This requires wallet connectivity and network access, which may surprise users who expect `--dry-run` to be offline. The error handling gracefully degrades (`nonce=(unknown)`), which is good, but the README doesn't mention this network requirement.

**Wallet initialized twice**: For the submission path, `WalletCore::from_env()` is called in Step 7 (nonce fetch) and again after the dry-run gate. The second initialization is redundant. Refactor to initialize once and reuse.

### 4. Security

**No concerns** — dry-run output doesn't expose private keys. Nonces are already observable on-chain. The privacy warning removal was correct (commit d2a3892).

### 5. Performance

**Double binary read** (noted above in correctness) — `fs::read(program_path)` happens twice on the submission path. Minor but avoidable.

---

### Summary of Required Changes

| Priority | Issue |
|----------|-------|
| 🔴 Bug | `--dry-run json` space syntax never matches — `json` arg is silently dropped |
| 🔴 Bug | Nonce `u128 as u64` truncation — silent data loss |
| 🟡 | Program binary loaded twice on submission path (TOCTOU + waste) |
| 🟡 | Transaction summary prints unconditionally (including non-dry-run) |
| 🟡 | `WalletCore` initialized twice |
| 🟡 | `raw_account_ids` mixed encoding (base58 vs hex) — needs follow-up issue |
| 💬 | Split rustfmt changes into separate commit for reviewability |
| 💬 | README should note `--dry-run` may contact the sequencer for nonces |

The feature design is sound and the JSON output structure is well-thought-out. The core issues are the dead-code `--dry-run json` parsing path and the nonce truncation. Once those are addressed and the submission-path double-read is cleaned up, this is good to merge.

---
🤖 Review by Claude Code

vpavlin (approved):

vpavlin (approved):
Looks good (when rebased)

PR #133, opened 2026-04-14
feat(spel-cli): show seed inputs in PDA derivation output

marclawclaw:
## Summary

Show seed inputs when displaying PDA derivation, so it's clear which seeds were used to derive each PDA.

## What changed

- `pda_seeds_json()`: builds a JSON array of seed type objects
- `format_pda_seeds()`: formats Const seeds as `"value"`, other seed types as `type: value`
- Dry-run output now shows seed inputs inline for each PDA

## Testing
- `cargo fmt --check` ✅
- `cargo check` ✅

## Related
Part of logos-co/spel#10 (wishlist)


marclawclaw (commented):
## Code Review — PR #133

**Verdict: REQUEST_CHANGES** (posted as comment since this is your own PR)

### Overview

This PR does significantly more than the title suggests. The title says "show seed inputs in PDA derivation output" but it actually:
1. Restructures dry-run to run **after** program loading, PDA resolution, and nonce fetching
2. Adds `--dry-run [text|json]` format support (collapsing two flags into one)
3. Shows PDA seed inputs inline — the stated goal
4. Applies `rustfmt` across ~25 files (the bulk of the diff)

The feature work (1–3) is solid. The `format_pda_seeds()` and `pda_seeds_json()` helpers are clean. Moving dry-run later in the pipeline so it shows the full picture is the right call. However, there are a few issues:

---

### Issues

#### 1. 🔴 `--dry-run " "` branch is dead code (`lib.rs` ~line 674)

```rust
s if s.starts_with("--dry-run ") => {
```

Shell argument splitting means a single `args[i]` will **never** contain `"--dry-run json"` as one string — the shell splits it into two args `["--dry-run", "json"]`. This branch is unreachable. Instead, the `--dry-run` branch should peek at `args[i+1]` for an optional format argument (similar to how `--idl` consumes its next arg).

#### 2. 🔴 `--dry-run` without format still prints "Dry run" footer even on non-dry-run (`tx.rs` ~line 491)

When `dry_run_format` is `None` (i.e., **not** a dry-run), the code still prints the full transaction summary box, signer nonces, and the "Dry run — this is what WOULD be submitted" message before checking `if dry_run_format.is_some() { return; }`. The summary and footer should be gated behind the dry-run check, not printed unconditionally.

#### 3. 🟡 Program binary read twice on submit path (`tx.rs` ~line 504)

The program binary is read and parsed in Step 4 (`_program_obj`), but `_program_obj` is unused — then it's read **again** after the dry-run check for the submit path. The underscore prefix signals intent but it's wasteful and confusing. Either use the first load or don't load until needed.

#### 4. 🟡 Nonce fetch error is silent in dry-run but fatal on submit

In Step 7, if `get_accounts_nonces` fails, dry-run shows `nonce=(unknown)` — reasonable. But if the user then removes `--dry-run`, the same error (`n.unwrap_or_else(|| ... exit(1))`) will crash with a confusing message about running `--dry-run` to see nonces. The nonce fetch should be retried or the error message should say "nonce fetch failed" rather than suggesting `--dry-run`.

#### 5. 🟡 `Nonce.0 as u64` cast (`tx.rs` ~line 420)

```rust
format!("nonce={}", n.0 as u64)
```

What is the underlying type of `Nonce.0`? If it's already `u64`, the cast is redundant. If it's larger, the cast silently truncates. Worth verifying.

#### 6. 🟢 Scope: rustfmt across 25 files

The formatting changes are fine individually but they inflate the diff from ~200 meaningful lines to 1766, making review harder. Consider splitting format-only changes into a separate commit or PR in the future.

---

### What's good

- `format_pda_seeds()` and `pda_seeds_json()` are clean, well-separated display/serialization helpers
- The dry-run restructuring (running after PDA resolution and nonce fetch) gives users the complete picture — this is the right design
- The `--dry-run [text|json]` collapse is a good UX improvement
- Box-drawing summary output is clear and informative
- JSON output structure is reasonable for scripting

### Summary

Fix items 1 and 2 (dead code and unconditional summary printing) before merge. Items 3–5 are worth addressing but not blocking.

PR #135, opened 2026-04-14
feat(spel-cli): show seed inputs in PDA derivation output

marclawclaw:
## Summary

Show seed inputs when displaying PDA derivation, so it's clear which seeds were used to derive each PDA.

When a PDA is resolved during dry-run or submit, the output now shows:

**Clean PR** — only the seed display feature, no dry-run infrastructure changes.

## What changed

- Added  function: formats Const seeds as , Account seeds as , Arg seeds as 
- Added  after each PDA derivation line

## Testing
- `cargo fmt --check` ✅
- `cargo check` ✅

## Related
Part of logos-co/spel#10 (wishlist)

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,19 @@ mod my_program {
 | `#[account(pda = arg("create_key"))]` | PDA derived from an instruction argument |
 | `members: Vec<AccountWithMetadata>` | Variable-length trailing account list |
 
+### PDA Seed Display
+
+When the CLI derives PDA accounts during transaction execution, it prints the seed inputs used for each derivation:
+
+```
+  PDA vault → 4Lp3gkH...
+    seeds: [program_id, "state"]
+  PDA token_account → 7xQ2m...
+    seeds: [program_id, Account(owner), Arg(create_key)]
+```
+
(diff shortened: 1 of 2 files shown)


marclawclaw (commented):
## Code Review — Five-Axis Assessment

### 1. Correctness ✅
`format_pda_seeds()` correctly maps all three `IdlSeed` variants to readable strings. Integration point is correctly placed inside the `Ok` arm after successful PDA computation. Prepending `"program_id"` is accurate.

### 2. Readability ⚠️ Minor
The function is placed **between** `use std::collections::HashMap;` and the remaining `use` statements. Rust convention is all imports first, then definitions. Consider moving the function below all `use` statements.

### 3. Architecture ✅
Display helper in `tx.rs` is the right place — it's presentation logic, not PDA computation.

### 4. Security ✅
Display-only output.

### 5. Performance ✅
No concerns.

---

**Note:** ~90% of the diff is rustfmt reformatting. The actual feature is ~10 lines. The formatting fixes are valid but bundling them makes the diff harder to review — consider separating formatting-only changes into their own commit in the future.

**Verdict: APPROVE** — clean, minimal feature addition. One nit on function placement (above imports).

PR #139, opened 2026-04-24
feat(spel-framework): support LEZ validity windows in program output

0x-r4bbit:
Add block and timestamp validity window support to SpelOutput so that SPEL programs can restrict the block or timestamp range in which their transaction is valid — a LEZ feature that was previously inaccessible.

Previously the macro-generated main() hardcoded unbounded validity windows in ProgramOutput regardless of what the program returned. SpelOutput now carries both windows and the macro threads them through, so a handler can express constraints like:

```
    Ok(SpelOutput::execute(accounts, calls)
        .with_block_validity_window(100..)
        .try_with_timestamp_validity_window(1_700_000_000..1_800_000_000)?)
```

Changes:
- spel-framework-core: add block_validity_window and timestamp_validity_window fields to SpelOutput; add four builder methods (with_block_validity_window, try_with_block_validity_window, with_timestamp_validity_window, try_with_timestamp_validity_window)
- spel-framework-core: export ValidityWindow, BlockValidityWindow, TimestampValidityWindow, InvalidWindow, BlockId, Timestamp from prelude so programs can construct windows without extra imports
- spel-framework-macros: extract windows from SpelOutput in match arms and pass them to ProgramOutput instead of hardcoding unbounded
- Existing programs are unaffected: all constructors default to ValidityWindow::new_unbounded(), preserving prior behaviour

### DIFF
--- a/spel-framework-core/src/lib.rs
+++ b/spel-framework-core/src/lib.rs
@@ -18,5 +18,9 @@ pub mod prelude {
     pub use crate::spel_output::AutoClaim;
     pub use crate::types::{IntoPostState, SpelOutput, AccountConstraint};
     pub use nssa_core::account::{Account, AccountWithMetadata};
-    pub use nssa_core::program::{AccountPostState, ChainedCall, Claim, PdaSeed, ProgramId};
+    pub use nssa_core::program::{
+        AccountPostState, BlockValidityWindow, ChainedCall, Claim, InvalidWindow, PdaSeed,
+        ProgramId, TimestampValidityWindow, ValidityWindow,
+    };
+    pub use nssa_core::{BlockId, Timestamp};
 }
--- a/spel-framework-core/src/spel_output.rs
+++ b/spel-framework-core/src/spel_output.rs
@@ -5,7 +5,7 @@
 //! `(Account, AutoClaim)` pairs into the correct `AccountPostState` values.
 
 use nssa_core::account::Account;
-use nssa_core::program::{AccountPostState, ChainedCall, Claim, PdaSeed};
+use nssa_core::program::{AccountPostState, ChainedCall, Claim, PdaSeed, ValidityWindow};
 
 use crate::types::{IntoPostState, SpelOutput};
 
@@ -128,6 +128,8 @@ impl SpelOutput {
         Self {
             post_states,
             chained_calls: calls,
+            block_validity_window: ValidityWindow::new_unbounded(),
+            timestamp_validity_window: ValidityWindow::new_unbounded(),
         }
     }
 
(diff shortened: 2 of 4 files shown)


fryorcraken:
Local review notes (non-blocking) — happy to land as-is or in a follow-up:

**Important**

1. **No test exercises the macro pass-through.** The 13 new unit tests cover the `SpelOutput` builders in isolation, but nothing drives the `#[lez_program]` macro with a non-default window. The wiring at `spel-framework-macros/src/lib.rs:269-329` and `:811` is mechanical, but a regression that hardcoded a window or dropped one would still compile cleanly. Suggest extending `tests/e2e/fixture_program` with one instruction that returns `SpelOutput::execute(...).with_block_validity_window(N..)` and asserts the value lands on the emitted `ProgramOutput`.

2. **`SpelOutput` has `pub` fields without `#[non_exhaustive]`.** Any external program constructing it via a struct literal breaks on this PR (and again on every future field). The struct already had public fields pre-PR, so this isn't introduced here, but this is the moment the pattern starts to bite. Recommend `#[non_exhaustive]` (or moving fields behind builders) so subsequent additions are non-breaking.

**Suggestions**

3. **Deprecating `into_parts` adds churn for no caller benefit** (`spel-framework-core/src/types.rs:104-108`). The only in-tree caller (the macro at lib.rs:811) is already migrated. Since fields are `pub`, downstream programs can just destructure. Either drop `into_parts` outright (pre-1.0) or rename `into_parts_with_windows` back to `into_parts`. Otherwise the deprecation warning will fire in user code that doesn't yet care about windows.

4. **The 4-tuple return from `into_parts_with_windows` will keep growing** (`types.rs:111-125`). Returning a struct — or just `SpelOutput` and letting the caller destructure — avoids touching every call site when a 5th field eventually joins. Cheap to fix now, expensive later.

6. **Test ergonomics.** The 13 new tests in `spel_output.rs:265-358` repeat `Vec::<(Account, AutoClaim)>::new()`. A one-line helper (`fn empty() -> SpelOutput { ... }`) makes each test scannable. Cosmetic.

7. **Coverage gap on the fallible path.** Behavior of `try_with_block_validity_window(..)` (i.e., `RangeFull` through `TryInto`) isn't asserted. Probably delegates to the same conversion as `Into`, but worth one test to lock the contract.

8. **Doc duplication** between `with_block_validity_window` and `try_with_block_validity_window` (`types.rs:61-62` and `:69-71`). Same summary line on both methods. Differentiate (e.g., "(infallible)" vs "(fallible — `Err` on empty range)") so rustdoc readers see the distinction without reading both bodies. Same comment applies to the timestamp pair.

Verified locally: `cargo build` (framework crates), `cargo test -p spel-framework-core` (41 passed, 13 new), and `cargo build` for `tests/e2e/fixture_program` all pass; no new clippy warnings.

PR #146, opened 2026-04-24
fix: unify IDL generation paths to include #[account_type] annotated types

### LINKED ISSUE #141: IDL generation paths diverge: make idl drops #[account_type] that spel generate-idl picks up
## Summary

SPEL has two ways to produce an IDL, and they disagree about \`#[account_type]\`:

| Path | How it's invoked | Picks up \`#[account_type]\` markers? |
|------|------------------|---------------------------------------|
| **CLI subcommand** | \`spel generate-idl methods/guest/src/bin/prog.rs\` (→ \`spel_framework_core::idl_gen::generate_idl_from_file\`) | ✅ yes, via \`collect_account_types(&file.items)\` |
| **Proc macro** | \`spel_framework::generate_idl!(\"…/prog.rs\")\` inside \`examples/src/bin/generate_idl.rs\`, invoked by \`make idl\` (→ \`spel_framework_macros::expand_generate_idl\` / \`generate_idl_json\`) | ❌ no |

The scaffold's \`Makefile\` wires \`make idl\` to the proc-macro path. So a user who follows the tutorial, writes \`#[account_type] pub struct CounterState { … }\` in the guest file, and runs \`make idl\`, gets an IDL with \`\"accounts\": []\`. Then \`spel inspect <PDA> --type CounterState\` silently can't find the type. Nothing in the output tells them why.

The CLI path produces the correct IDL (I confirmed this during the PR #138 tutorial walk):

\`\`\`bash
spel generate-idl methods/guest/src/bin/my_counter.rs | jq .accounts
# [ { \"name\": \"CounterState\", \"type\": { \"kind\": \"struct\", \"fields\": [ … ] } } ]
\`\`\`

## Why this happens

\`spel_framework_core/src/idl_gen.rs\` (CLI path) calls \`collect_account_types(&file.items)\` after parsing the program source. \`spel_framework_macros/src/lib.rs::generate_idl_json\` (proc-macro path) walks \`instructions\` but never iterates \`file.items\` looking for \`#[account_type]\` markers — so the top-level structs/enums annotated with it are silently dropped when emitting \`PROGRAM_IDL_JSON\`.

## Proposed fix

Two options:

1. **Unify on one path.** Have \`generate_idl_fn\` / \`generate_idl_json\` in \`spel-framework-macros\` delegate to the same \`collect_account_types\` logic (or to \`idl_gen::generate_idl_from_str\` directly). Both produce an \`SpelIdl\`; the proc-macro path just needs to serialize it.

2. **Port the collection.** Copy the \`collect_account_types\` logic into \`generate_idl_json\` so it builds the \`accounts\`/\`types\` entries from the file's top-level items. Less invasive but duplicates code.

Option 1 is the clean fix.

## Workaround (until fixed)

Replace \`make idl\` with \`spel generate-idl methods/guest/src/bin/<name>.rs > <name>-idl.json\`. The tutorial in PR #138 tells users to do this and explains why.

## Impact

\`spel inspect --type …\` is the primary way to read account data off-chain, and it's prominently advertised in the README (\"inspect subcommand to extract ProgramId from binaries and decode account data\"). Any user who defines state via \`#[account_type]\` and generates the IDL via \`make idl\` will hit this.

Found during the tutorial walk that produced #138.

cc @fryorcraken

vpavlin:
## Summary

Fixes [#141](https://github.com/logos-co/spel/issues/141) — The proc-macro path for IDL generation was hardcoding empty `"accounts"` and `"types"` arrays, causing a divergence with the CLI path (`spel generate-idl`) which correctly collects `#[account_type]`-annotated types.

### New file: `spel-framework-macros/src/account_types.rs`
Shared module containing `collect_account_types()` and all helper functions for scanning Rust source items for `#[account_type]`-annotated types and resolving their transitive type references.

### Modified: `spel-framework-macros/Cargo.toml`
- Added `serde_json = "1.0"` dependency (for serializing account/type data at macro expansion time)
- Added `spel-framework-core` with `idl-gen` feature (provides IDL type definitions)

### Modified: `spel-framework-macros/src/lib.rs`
- **`generate_idl_fn`**: Now accepts `accounts` and `types` parameters, serializes them to JSON at expansion time, and embeds them in generated code for runtime deserialization
- **`generate_idl_json`**: Now accepts `accounts` and `types` parameters and embeds their JSON representation directly into the IDL string
- **`expand_lez_program`**: Reads the source file via `CARGO_MANIFEST_DIR`, parses it, and passes top-level items to `collect_account_types()`
- **`expand_generate_idl`**: Already had access to `file.items`; now passes them to `collect_account_types()`

## Testing
- All 87 workspace tests pass
- E2E IDL generation test passes
- CLI IDL generation tests pass

## Impact
Users who define `#[account_type]` structs/enums and generate IDL via `make idl` (proc-macro path) will now get correct account type definitions in their IDL, matching the output of `spel generate-idl` (CLI path). This fixes the silent mismatch that caused `spel inspect --type …` to fail finding types.

### DIFF
--- a/spel-framework-macros/Cargo.toml
+++ b/spel-framework-macros/Cargo.toml
@@ -8,7 +8,9 @@ description = "Proc macros for the SPEL program framework"
 proc-macro = true
 
 [dependencies]
-proc-macro2 = "1.0"
+proc-macro2 = { version = "1.0", features = ["span-locations"] }
 quote = "1.0"
 syn = { version = "2.0", features = ["full", "extra-traits", "visit-mut"] }
 sha2 = "0.10"
+serde_json = "1.0"
+spel-framework-core = { path = "../spel-framework-core", features = ["idl-gen"] }
--- a/spel-framework-macros/src/account_types.rs
+++ b/spel-framework-macros/src/account_types.rs
@@ -1,0 +1,257 @@
+//! Shared account type scanning logic for IDL generation.
+//!
+//! This module provides functions to scan Rust source items for `#[account_type]`-annotated
+//! types and collect helper types referenced by them. Both the CLI path (`spel generate-idl`)
+//! in `spel-framework-core` and the proc-macro path (`lez_program`, `generate_idl!`) use this
+//! logic to ensure consistent IDL output.
+
+use std::collections::HashSet;
+
+use syn::{Attribute, Item, ItemEnum, ItemStruct, Type};
+
+use spel_framework_core::idl::{
+    IdlAccountType, IdlEnumVariant, IdlField, IdlType, IdlTypeDef,
+};
+
+// ─── Account type scanning ────────────────────────────────────────────────
+
+/// Check if an item has the `#[account_type]` attribute.
+pub fn has_account_type_attr(attrs: &[Attribute]) -> bool {
+    attrs.iter().any(|a| a.path().is_ident("account_type"))
+}
+
+/// Convert a Rust `syn::Type` to an IDL type representation.
+pub fn syn_type_to_idl_type(ty: &Type) -> IdlType {
+    match ty {
+        Type::Path(type_path) => {
+            let segment = match type_path.path.segments.last() {
+                Some(s) => s,
+                None => return IdlType::Primitive("unknown".to_string()),
+            };
+            let ident = segment.ident.to_string();
+            match ident.as_str() {
+                "u8" | "u16" | "u32" | "u64" | "u128" | "i8" | "i16" | "i32" | "i64"
+                | "i128" | "bool" | "String" => IdlType::Primitive(ident.to_lowercase()),
+                "Vec" => {
+                    if let syn::PathArguments::AngleBracketed(args) = &segment.arguments {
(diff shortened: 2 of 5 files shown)


vpavlin:
Review notes — three things to address before merging:

**1. Silent failure when source file not found**
When the proc-macro can't locate `src/bin/{module_name}.rs` it silently falls back to empty `accounts`/`types`. The IDL looks valid but is missing account types with no signal to the developer. Should emit `compile_error!` or a warning.

**2. Hardcoded path convention is undocumented**
The assumption that modules live at `$CARGO_MANIFEST_DIR/src/bin/{name}.rs` should be documented in a comment near the derivation logic.

**3. No tests for the proc-macro path**
The CLI path has good coverage but the new macro path has none. A test asserting that `#[lez_program]` on a module with `#[account_type]` structs produces the correct `accounts` in the IDL would catch regressions.

Core unification logic is sound — BFS collection and the shared module structure look good.

vpavlin:

vpavlin (commented):
All Copilot review comments addressed. Please re-review.

PR #147, opened 2026-04-24
feat(spel-client-gen): generate C FFI fetch functions for account types with PDAs

vpavlin:
## Summary

`spel-client-gen` now generates C FFI stub functions for **reading account state**, in addition to the existing transaction submission functions. This closes the gap where the C FFI was write-only — useful for submitting transactions but useless for building reactive UIs.

### New: `generate_account_fetch_functions()`
- Collects account types with PDAs across all instructions (deduplicated by name)
- Generates one `extern "C"` fetch function per unique PDA account
- Each function:
  1. Parses JSON args (`wallet_path`, `sequencer_url`, `program_id_hex`, plus any PDA seed args)
  2. Computes the PDA using the existing `compute_{account}_pda()` helper
  3. Fetches the account from the sequencer via `wallet.sequencer_client.get_account()`
  4. Borsh-decodes the raw bytes into the account type
  5. Serialises the decoded state as a JSON object and returns it

### Updated: `generate_header()`
- Adds `char* {prefix}_fetch_{account_name}(const char* args_json);` declarations for each fetch function

### Updated: `generate_ffi()`
- Conditionally imports `use borsh::BorshDeserialize;` when fetch functions are generated

### Updated: `seed_arg_codegen()` (codegen.rs)
- Changed visibility from `fn` to `pub(crate)` so it can be reused in FFI codegen

## Example Output

For a program with a `WhisperState` account type and PDA, the generated C header now includes:

```c
/* Fetch WhisperState account state */
char* whisper_wall_fetch_whisper_state(const char* args_json);
```

The function accepts JSON with `wallet_path`, `sequencer_url`, `program_id_hex` (and any additional seed args), and returns:

```json
{
  "success": true,
  "state": {
    "admin": "hex_encoded_bytes",
    "latest_whisper": "...",
    "last_tip": "0",
    "whisper_count": 0,
    "total_tips": 0
  }
}
```

Or on error:
```json
{
  "success": false,
  "error": "fetch WhisperState: ..."
}
```

## Testing
- All 24 existing tests pass (no regressions)
- 4 new tests added:
  - `test_ffi_fetch_function_generation` — verifies fetch function in FFI + header for whisper-wall-like IDL
  - `test_ffi_fetch_with_arg_seed` — verifies arg-based PDA seeds work correctly
  - `test_ffi_no_fetch_without_accounts` — ensures no fetch functions when no PDAs
  - `test_fetch_deduplication` — verifies same account type across multiple instructions gets only one fetch function

### DIFF
--- a/spel-client-gen/src/codegen.rs
+++ b/spel-client-gen/src/codegen.rs
@@ -265,7 +265,7 @@ fn collect_pda_helpers(idl: &SpelIdl) -> Vec<PdaHelper> {
 
 /// Generate codegen expressions for a seed argument based on its Rust type.
 /// Returns (param_type, optional_let_binding, seed_expression).
-fn seed_arg_codegen(name: &str, rust_type: &str) -> (String, Option<String>, String) {
+pub(crate) fn seed_arg_codegen(name: &str, rust_type: &str) -> (String, Option<String>, String) {
     match rust_type {
         "AccountId" => (
             "&AccountId".to_string(),
--- a/spel-client-gen/src/codegen.rs
+++ b/spel-client-gen/src/codegen.rs
@@ -275,7 +275,7 @@ pub(crate) fn seed_arg_codegen(name: &str, rust_type: &str) -> (String, Option<S
         "[u8; 32]" | "[u8;32]" => (
             format!("&{}", rust_type),
             None,
-            format!("{}", name),
+            name.to_string(),
         ),
         "ProgramId" | "[u32; 8]" | "[u32;8]" => (
             "&ProgramId".to_string(),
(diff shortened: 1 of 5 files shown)


vpavlin:
Review notes — one blocker plus minor items:

**Blocker: JSON parsing bug**
The generated fetch functions appear to parse an empty object instead of `args_json`:
```rust
serde_json::from_value(serde_json::json!({}))
```
This should be `serde_json::from_str(args_json)`. As-is, all generated fetch functions will fail at runtime because they're never reading the actual arguments passed in.

**Minor: generic error messages for missing seed args**
When a required seed argument is missing, the error doesn't identify which argument, making debugging harder for users.

**Minor: no test for error paths**
Only happy-path coverage. A test that omits a required seed arg and verifies a useful error is returned would be good.

The overall design is solid — deduplication logic, PDA computation reuse, and the new tests for the happy path are all well done. Just needs the JSON parsing bug fixed before merge.

vpavlin:

vpavlin:
This got a bit too off-topic

PR #148, opened 2026-04-24
fix: map lowercase 'string' IDL type to Rust 'String'

### LINKED ISSUE #142: spel-client-gen: emits lowercase `string` instead of `String` in generated instruction enum variants
## Summary

`spel-client-gen` generates invalid Rust code: instruction enum variants that have `String`-typed arguments are emitted with `string` (lowercase) instead of `String`. The generated file does not compile.

## Reproduction

```bash
# Given any SPEL program with a String instruction argument, e.g.:
spel-client-gen --idl whisper-wall-idl.json --out-dir /tmp/out
```

`/tmp/out/whisper_wall_ffi.rs` (and `whisper_wall_client.rs`) contain:

```rust
pub enum WhisperWallInstruction {
    Initialize,
    Whisper {
        msg: string,   // ← should be String
    },
    Overwrite {
        msg: string,   // ← should be String
        tip: u128,
    },
    DrainJar,
    Reveal,
}
```

`string` is not a valid Rust type; `cargo build` fails immediately with `cannot find type \`string\` in this scope`.

The same bug appears in the async client (`whisper_wall_client.rs`) method signatures:

```rust
pub async fn whisper(&self, accounts: WhisperAccounts, msg: string) -> Result<String, String> {
```

## IDL fragment that triggers it

```json
{
  "name": "whisper",
  "args": [{ "name": "msg", "type": "string" }]
}
```

The IDL type name `"string"` (JSON lowercase) is being copied verbatim into the Rust output instead of being mapped to `String`.

## Expected behaviour

The IDL type `"string"` should map to Rust `String`. Every other primitive mapping works correctly (`u128` → `u128`, `u64` → `u64`, etc.).

## Fix hint

In `spel-client-gen`'s type-mapping function, add a case for `"string"` → `"String"` alongside the existing numeric mappings.

vpavlin:
## Summary

Fixes #142 - `spel-client-gen` generates invalid Rust code when an IDL argument has type `"string"` (lowercase). The generated file does not compile because `string` is not a valid Rust type.

## Changes

- Add `"string" => "String"` mapping in `idl_type_to_rust()` to fix client code generation
- Add `"string" | "String"` handling in `idl_type_to_json_parse()` for FFI code generation
- Add test case `test_string_type_lowercased` to verify the fix

## Test plan

- [x] All 21 existing tests pass
- [x] New test `test_string_type_lowercased` passes

### DIFF
--- a/spel-client-gen/src/tests.rs
+++ b/spel-client-gen/src/tests.rs
@@ -694,3 +694,33 @@ fn test_no_pda_no_helpers() {
         "should not generate fetch helpers when no PDAs"
     );
 }
+
+#[test]
+fn test_string_type_lowercased() {
+    let idl = r#"{
+        "version": "0.1.0",
+        "name": "whisper_wall",
+        "instructions": [{
+            "name": "whisper",
+            "accounts": [
+                {"name": "user", "writable": true, "signer": true, "init": false}
+            ],
+            "args": [{"name": "msg", "type": "string"}]
+        }]
+    }"#;
+    let output = generate_from_idl_json(idl).expect("codegen should succeed");
+    
+    // Client code should use `String` (uppercase), not `string` (lowercase)
+    assert!(
+        output.client_code.contains("msg: String"),
+        "client code should have msg: String, got:\n{}",
+        output.client_code
+    );
+    
+    // FFI code should also use `String`
+    assert!(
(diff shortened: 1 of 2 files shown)


vpavlin:

PR #149, opened 2026-04-24
fix: strip Public/Private prefix in generated parse_account_id

### LINKED ISSUE #144: spel-client-gen: generated parse_account_id does not accept wallet Public/Private prefix format
## Summary

The generated FFI code calls `.parse()` directly on the account ID string passed in from the JSON args. However, the `wallet account list` command outputs account IDs in `Public/<base58>` / `Private/<base58>` format (e.g. `Public/6iArKUXxhUJqS7kCaPNhwMWt3ro71PDyBj7jwAyE2VQV`), and `AccountId::from_str` does not understand this prefix — it only accepts the raw base58 key.

This means any user who copies an account ID from `wallet account list` and passes it to a generated FFI function gets:

```
invalid AccountId: Public/6iArKUXxhUJqS7kCaPNhwMWt3ro71PDyBj7jwAyE2VQV
```

## Expected behaviour

The generated `parse_account_id` helper should strip the `Public/` or `Private/` prefix before parsing, consistent with how the wallet CLI itself handles this format (see `wallet/src/helperfunctions.rs::parse_addr_with_privacy_prefix`).

## Suggested fix

```rust
fn parse_account_id(s: &str) -> Result<AccountId, String> {
    let base58 = s
        .strip_prefix("Public/")
        .or_else(|| s.strip_prefix("Private/"))
        .unwrap_or(s);
    base58.parse().map_err(|_| format!("invalid AccountId: {}", s))
}
```

## Context

Discovered while building the WhisperWall SPEL program as a Basecamp UI plugin and testing via the plugin UI, where account IDs are entered by copy-pasting from `wallet account list` output.

vpavlin:
## Summary

The generated FFI code's `parse_account_id` function now strips `Public/` and `Private/` prefixes before parsing, consistent with how the wallet CLI outputs account IDs (e.g. `wallet account list`).

This fixes the issue where users who copy-paste account IDs from `wallet account list` get an `invalid AccountId` error.

## Changes

- **`spel-client-gen/src/ffi_codegen.rs`**: Added prefix stripping in the generated `parse_account_id` function
- **`spel-client-gen/src/tests.rs`**: Added test verifying the prefix-stripping behavior

Fixes #144

### DIFF
--- a/spel-client-gen/src/ffi_codegen.rs
+++ b/spel-client-gen/src/ffi_codegen.rs
@@ -47,7 +47,6 @@ pub fn generate_ffi(idl: &SpelIdl) -> Result<String, String> {
     writeln!(out, "use nssa::public_transaction::{{Message, WitnessSet}};").unwrap();
     writeln!(out, "use sequencer_service_rpc::RpcClient as _;").unwrap();
     writeln!(out, "use wallet::WalletCore;").unwrap();
-    writeln!(out, "use std::panic::UnwindSafe;").unwrap();
 
     // Import or generate instruction type
     if let Some(ref itype) = idl.instruction_type {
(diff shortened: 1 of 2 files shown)


vpavlin:

PR #150, opened 2026-04-24
fix: wrap generated extern "C" FFI functions with catch_unwind

vpavlin:
## Summary

All generated `extern "C"` FFI entry points now wrap implementation calls
with `std::panic::catch_unwind` to prevent FFI panics from aborting the
host process.

### Problem

The generated `extern "C"` FFI functions did not catch Rust panics. If any
code path inside the implementation panics (including panics inside `nssa` /
`wallet` crate internals — e.g. `execute_and_prove(...).unwrap()` in the
privacy-preserving TX path), the panic attempts to unwind through C++ stack
frames that are not unwind-safe, causing the entire host process to abort
immediately.

This is undefined behaviour at the FFI boundary and is explicitly warned against
in the Rust reference.

### Fix

- Added `use std::panic::UnwindSafe;` import to generated FFI code
- Added `ffi_call()` helper that wraps with `std::panic::catch_unwind`:
  - `Ok(Ok(r))` → returns success JSON
  - `Ok(Err(e))` → returns error JSON
  - `Err(_) (panic)` → returns `{"success":false,"error":"internal panic in FFI"}`
- Replaced direct `match {fn_name}_impl(args)` with `ffi_call(move || {fn_name}_impl(args))`

### Testing

- All 20 existing tests pass
- Added `test_ffi_has_catch_unwind` test to verify panic-safety in generated code

### DIFF
--- a/spel-client-gen/src/ffi_codegen.rs
+++ b/spel-client-gen/src/ffi_codegen.rs
@@ -102,7 +102,12 @@ pub fn generate_ffi(idl: &SpelIdl) -> Result<String, String> {
     writeln!(out, "    match std::panic::catch_unwind(f) {{").unwrap();
     writeln!(out, "        Ok(Ok(r))  => to_cstring(r),").unwrap();
     writeln!(out, "        Ok(Err(e)) => error_json(&e),").unwrap();
-    writeln!(out, "        Err(_)     => error_json(\"internal panic in FFI\"),").unwrap();
+    writeln!(out, "        Err(e) => {{").unwrap();
+    writeln!(out, "            let msg = e.downcast_ref::<&str>().copied()").unwrap();
+    writeln!(out, "                .or_else(|| e.downcast_ref::<String>().map(|s| s.as_str()))").unwrap();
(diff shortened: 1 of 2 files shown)


vpavlin:
Review notes — two things worth addressing:

**1. Panic message is discarded**
The current implementation catches the panic but returns a generic `"internal panic in FFI"`, throwing away the actual panic payload. Users (and developers) get no useful debugging info. Should downcast to extract the message:
```rust
Err(e) => {
    let msg = e.downcast_ref::<&str>().copied()
        .or_else(|| e.downcast_ref::<String>().map(|s| s.as_str()))
        .unwrap_or("<unknown panic>");
    error_json(&format!("panic: {}", msg))
}
```

**2. `version()` not wrapped (minor)**
`version()` calls `to_cstring()` which can technically panic on null bytes. For consistency, either wrap it too or add a comment explaining why it's safe to leave bare.

The core fix (preventing host process abort when FFI panics) is correct and covers the critical instruction entry points. These are improvements rather than blockers, but #1 especially would make this much more useful in practice.

vpavlin:

PR #174, opened 2026-05-05
feat: extend private PDA support to CLI dispatch and smoke test

vpavlin:
## Summary

Follow-up to #171 (private PDA macro support). Completes the end-to-end story for private PDAs:

- **`IdlPda.npk_arg`**: adds `npk_arg: Option<String>` to the IDL PDA struct so the CLI knows which instruction arg supplies the NullifierPublicKey — without it, the CLI couldn't compute private PDA addresses at transaction time
- **CLI dispatch fix**: `tx.rs` was passing `None` for npk when computing PDAs, always producing a public address; now reads the correct arg from the IDL
- **`NullifierPublicKey` in CLI args**: `parse.rs` and `serialize.rs` now handle `IdlType::Defined{"NullifierPublicKey"}` (hex → 32-byte tuple) so instructions that take an npk arg work end-to-end
- **Smoke test**: adds `init_private_pda` instruction to the test program and steps 11–13: retrieve npk from wallet, compute private PDA address via `spel pda`, submit the instruction

## Test plan

- [ ] All workspace unit tests pass (`cargo test --workspace`)
- [ ] Fixture program tests pass including `idl_init_private_account_marks_pda_as_private` (now also asserts `npk_arg == "user_npk"`)
- [ ] Smoke test: `LEZ_TAG=v0.2.0-rc3 LSSA_DIR=<path> ./scripts/smoke-test-privacy.sh` completes all 13 steps including the new private PDA TX


vpavlin:
## Step 13 (private PDA TX) blocked on LEZ

The smoke test passes through step 12 (PDA address computation) but step 13 (`init-private-pda` as a privacy-preserving transaction) is currently skipped because it requires a LEZ change that is not in v0.2.0-rc3.

**Root cause**: The privacy-preserving circuit validates private PDA claims using visibility mask=3, which calls `AccountId::for_private_pda(program_id, seed, npk)` to verify the account address. The wallet's `PrivacyPreservingAccount` enum in rc3 has no variant that produces mask=3 — it only has `Public` (mask=0), `PrivateOwned` (mask=1/2), and `PrivateForeign` (mask=2). There is no way to reach mask=3 from outside the LEZ crate since `key_protocol` (needed to compute the SSK/EPK) is an internal dependency not exposed to downstream crates.

**Where it's missing in rc3**: [`wallet/src/privacy_preserving_tx.rs@v0.2.0-rc3#L13`](https://github.com/logos-blockchain/logos-execution-zone/blob/v0.2.0-rc3/wallet/src/privacy_preserving_tx.rs#L13) — the `PrivacyPreservingAccount` enum has three variants, no `PrivatePda`.

**Where it exists in main**: [`wallet/src/privacy_preserving_tx.rs@main#L26`](https://github.com/logos-blockchain/logos-execution-zone/blob/main/wallet/src/privacy_preserving_tx.rs#L26) — `PrivatePda { nsk, npk, vpk, program_id, seed }` is present, along with `private_pda_preparation()` that computes the correct `for_private_pda` account ID and `account_identities()` that emits `InputAccountIdentity::PrivatePdaInit`.

**Unblocking**: bump the LEZ dependency from `v0.2.0-rc3` to a tag/commit that includes the `PrivatePda` variant. Once that lands, the fix in the SPEL CLI is a one-liner in `tx.rs` where the TODO comment is — replace `Public(id)` with `PrivatePda { nsk, npk, vpk, program_id, seed }` for private PDAs, and re-enable step 13 in the smoke test.

PR #182, opened 2026-05-06
feat: expose execution context to instruction handlers (issue #172)

vpavlin:
## Summary

Instruction handlers can now access trusted execution metadata (`self_program_id`, `caller_program_id`) through a `ProgramContext` parameter, without polluting the instruction ABI or IDL.

### New Features

**1. `ProgramContext` parameter injection**

```rust
#[instruction]
pub fn initialize(
    ctx: ProgramContext,
    #[account(owner = self_program_id)]
    definition: AccountWithMetadata,
) -> SpelResult {
    // ctx.self_program_id is the currently executing program
    // ctx.caller_program_id is the calling program
}
```

The macro-generated dispatcher injects `ProgramContext::new(self_program_id, caller_program_id)` at call time. The context parameter:
- Does **not** appear in the IDL
- Does **not** appear in generated instruction enums
- Is always passed as the first argument (matching typical function signatures)

**2. `#[account(owner = expr)]` constraint validation**

```rust
#[instruction]
pub fn initialize(
    #[account(owner = self_program_id)]
    definition: AccountWithMetadata,
) -> SpelResult { ... }
```

Generates validation that checks `account.program_owner == self_program_id`. The owner constraint is also reflected in the IDL output.

### Changes

- **spel-framework-core/src/context.rs** — New `ProgramContext` struct with `self_program_id` and `caller_program_id`
- **spel-framework-core/src/error.rs** — New `AccountOwnerMismatch` error variant (code 1010)
- **spel-framework-macros/src/lib.rs** — Detect context params, inject at dispatch, generate owner validation, update IDL output

### Tests

- 6 new tests for `ProgramContext` construction and usage patterns
- 4 new tests for owner constraint validation (match, mismatch, ordering, error code)
- All 230 existing + new tests pass

### Backward Compatibility

Fully backward compatible — no breaking changes. Existing instructions without context parameters work identically.

PR #188, opened 2026-05-08
feat(macros): support #[self_program_id] parameter injection in handlers

0x-r4bbit:
Handlers inside #[lez_program] modules can now declare a parameter annotated with #[self_program_id] to receive the program's own ID at runtime. The macro detects the attribute, excludes the parameter from the generated Instruction enum and IDL args, and injects the `self_program_id` value (already available in the generated main()) as the last positional argument in the dispatch match arm.

The attribute name is fixed; the parameter name is flexible. The parameter must appear last in the handler signature due to the positional call-arg construction.


0x-r4bbit:
This will likely be superseded by https://github.com/logos-co/spel/pull/182

Closing this one.

PR #189, opened 2026-05-11
feat(cli): parse and serialize Vec<String> args

Thompsonmina:
## Problem

spel-cli cannot parse `Vec<String>` arguments from the command line. The macro already emits the correct IDL shape (`Vec { vec: Primitive("string") }`) for any guest instruction parameter typed `Vec<String>`, and `DynamicValue::Str` already exists, so single-string args work today. But `parse_vec` in [`spel-cli/src/parse.rs`](spel-cli/src/parse.rs) has no arm for `Primitive("string")`, so the input falls through to `ParsedValue::Raw` and `to_dynamic_value` in [`spel-cli/src/serialize.rs`](spel-cli/src/serialize.rs) rejects it:

```
❌ Serialization error: type mismatch:
   expected Vec { vec: Primitive("string") },
   got Raw("bafybeiTestSingleCid000000000000000000000001")
```

This blocks any LEZ program whose instruction takes a list of identifiers, names, or other string-keyed inputs.

## Solution

Three small additive changes — no existing behaviour is altered:

1. **`spel-cli/src/parse.rs`** — add a `StringVec(Vec<String>)` variant to `ParsedValue`, plus the matching `Display` arm and a new arm in `parse_vec` for `Primitive("string")` / `Primitive("String")`. Format is comma-separated, mirroring how `Vec<u8>`, `Vec<u32>`, and `Vec<[u8; 32]>` are already specified.
2. **`spel-cli/src/serialize.rs`** — add an arm to `to_dynamic_value` that turns `ParsedValue::StringVec(strs)` into `DynamicValue::Seq(strs.map(DynamicValue::Str))`. `DynamicValue::Str` already serialises to risc0 serde via `serialize_str`, so the wire format follows automatically.
3. **`spel-cli/src/cli.rs`** — one extra line in the TYPE FORMATS help block.

## Verification

```bash
cargo test -p spel
# ... 83 passed; 0 failed
```

New tests:
- `parse::tests::vec_string_parses_csv` — CSV → `StringVec`.
- `parse::tests::vec_string_empty_input_yields_empty_vec`.
- `parse::tests::vec_string_single_element_yields_singleton`.
- `serialize::tests::to_dynamic_value_vec_string_emits_seq_of_str` — `StringVec` → `DynamicValue::Seq[Str, …]`.
- `serialize::tests::serde_roundtrip_vec_string` — **the critical contract test**: CLI-emitted bytes deserialise back through risc0's `Deserializer` as `Vec<String>`, the same code the guest runs.

## Known limitations

Elements containing commas are not supported (CSV format limitation). A follow-up PR can add quoting for general-purpose string vectors.

## Checklist

- [x] Builds cleanly (`cargo build`)
- [x] Tests pass (`cargo test -p spel` — 83 passing)
- [x] README updated — N/A (no user-facing surface change beyond one help-text line, included in this PR)
- [x] New public methods have doc comments — N/A (no new public items)
- [x] Branch is off `main`

---

## Update — flag repetition (per review thread)

Switched from CSV (`--cids a,b,c`) to flag repetition (`--cid a --cid b --cid c`) per the
discussion above. The on-the-wire format is unchanged — `ParsedValue::StringVec` →
`DynamicValue::Seq(Str…)` → risc0 serde, same bytes as before. What changed is the entry
path:

- `parse_instruction_args` now returns `HashMap<String, Vec<String>>`. Each `--key value`
  appends; scalar callers read `.last()` (same last-write-wins semantics as before), and
  `Vec<String>` args consume the whole vec.
- New `parse_string_vec(values: &[String]) -> ParsedValue` is the typed entry point for
  repeated flags. The CSV arm for `Primitive("string"|"String")` is removed from
  `parse_vec`; the `StringVec` variant and its `Display`/serialize paths are unchanged.
- `tx.rs` dispatches on `Vec<String>` shape at the parse call site (`is_vec_string`
  helper); every other type goes through the unchanged scalar `parse_value` path.

### Test changes

The three CSV-format unit tests in `parse.rs` (`vec_string_parses_csv`,
`vec_string_empty_input_yields_empty_vec`, `vec_string_single_element_yields_singleton`)
are replaced with four repetition-flow tests:

- `parse_string_vec_multiple_values`
- `parse_string_vec_empty_input_yields_empty_vec`
- `parse_string_vec_single_element_yields_singleton`
- `parse_string_vec_preserves_commas_in_elements`

`to_dynamic_value_vec_string_emits_seq_of_str` and `serde_roundtrip_vec_string` in
`serialize.rs` updated to construct input via `parse_string_vec(&[…])` instead of
`parse_value("foo,bar,baz", …)`. Wire-format assertions unchanged.

```bash
cargo test -p spel
# 84 passed; 0 failed
```

### DIFF
--- a/spel-cli/src/cli.rs
+++ b/spel-cli/src/cli.rs
@@ -61,6 +61,7 @@ pub fn print_help(idl: &SpelIdl, binary_name: &str) {
     println!("  [u8; N]               Hex string (2*N hex chars) or UTF-8 string (≤N chars, right-padded)");
     println!("  [u32; 8] / program_id Comma-separated u32s: \"0,0,0,0,0,0,0,0\"");
     println!("  Vec<[u8; 32]>         Comma-separated hex strings: \"aabb...00,ccdd...00\"");
+    println!("  Vec<String>           Repeat the flag, one element per occurrence: --foo a --foo b --foo c");
     println!();
     println!("CONFIG:");
     println!("  Create a spel.toml in your project root to avoid passing --idl and --program:");
@@ -126,18 +127,40 @@ pub fn print_instruction_help(ix: &IdlInstruction) {
     }
 }
 
-/// Parse CLI args for an instruction into a key-value map.
-pub fn parse_instruction_args(args: &[String], ix: &IdlInstruction) -> HashMap<String, String> {
-    let mut map = HashMap::new();
+/// Parse CLI args for an instruction into a key → list-of-values map.
+///
+/// Each `--key value` occurrence is appended to the entry for `key`, so a
+/// flag may be repeated.  Scalar args take the last value; `Vec<String>`
+/// args consume every value supplied.  A bare `--flag` with no value is
+/// only legal for boolean args (per the IDL) or the universal `--help`/
+/// `-h`; for anything else we exit with a `missing value` error so a
+/// missing CLI value can't be silently stored as the literal "true".
+pub fn parse_instruction_args(
+    args: &[String],
+    ix: &IdlInstruction,
+) -> HashMap<String, Vec<String>> {
+    let mut map: HashMap<String, Vec<String>> = HashMap::new();
     let mut i = 0;
     while i < args.len() {
         if args[i].starts_with("--") {
             let key = args[i][2..].to_string();
             if i + 1 < args.len() && !args[i + 1].starts_with("--") {
-                map.insert(key, args[i + 1].clone());
+                map.entry(key).or_default().push(args[i + 1].clone());
                 i += 2;
             } else {
-                map.insert(key, "true".to_string());
+                // Bare flag (no value follows).  Only valid for bool args
+                // or --help/-h; everything else is a user error.
+                let is_bool = ix.args.iter().any(|a| {
+                    snake_to_kebab(&a.name) == key
+                        && matches!(&a.type_, IdlType::Primitive(p) if p == "bool")
+                });
+                let is_help = key == "help" || key == "h";
+                if is_bool || is_help {
+                    map.entry(key).or_default().push("true".to_string());
+                } else {
+                    eprintln!("❌ --{}: missing value", key);
+                    std::process::exit(1);
+                }
                 i += 1;
             }
         } else {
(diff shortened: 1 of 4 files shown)


vpavlin:
Question is - is it better to have comma-separated list in a single cli flag, or figure out and allow repetition? 

```
--numbers 1,2,3
```

vs.

```
--number 1 --number 2 --number 3
```

The latter is more verbose a feels a bit cleaner

Thoughs @Thompsonmina @0x-r4bbit ?

0x-r4bbit:
Prefer the latter as well.

Thompsonmina:
yeah, i agree latter is a bit clearer intention wise, although i personally prefer former as its a bit more succinct . But the latter also avoids any clunkiness with commas so i agree that we should probably go with that.

vpavlin:
Hey @Thompsonmina — the direction is clear from the thread: repeated flags (`--number 1 --number 2 --number 3`) rather than CSV. Could you update the PR to use that approach? Once updated, CI should be quick to re-run and we can merge.

Thompsonmina:
Done! should probably rebase onto the new main right?

vpavlin:
Yeah, this seems to have diverged quite a bit:)

spel-cli/src/cli.rs
```diff
@@ -20,8 +20,7 @@ pub fn print_help(idl: &SpelIdl, binary_name: &str) {
     println!("  --bin-<NAME> <FILE>        Additional program binary (auto-fills --<NAME>-program-id)");
     println!();
     println!("COMMANDS:");
-    println!("  program-id <FILE> [--format hex|json]  Extract ProgramId from ELF binary(ies)");
```
vpavlin, line 31: This is wrong - we want to keep the program-id flag:)

Thompsonmina: Good catch, wasn't intentional. When I rebased onto current main must have carried that from the old head. Will fix now

weboko:
browsed through and looks good to me
@vpavlin anything blocking on your side?

weboko (approved):

weboko:
triggered tests, let's see if CI passes (cc @vpavlin )

Issue #3, opened 2026-02-20
Variable-count accounts: support dynamic account lists (e.g. Vec<Account> in instructions)


jimmy-claw:
## Problem

The `#[nssa_program]` macro assumes a **fixed** number of `AccountWithMetadata` parameters per instruction. This breaks down for programs that need a variable number of accounts at runtime — most notably the `lez-multisig` `CreateMultisig` and `Execute` instructions.

### `CreateMultisig`

Expects accounts:
- `accounts[0]`: `multisig_state` (PDA)
- `accounts[1..N+1]`: `N` member accounts (variable, determined by `members` arg)

With the current framework, you can only declare a fixed list of `AccountWithMetadata` params. There's no way to express "one account per member in the `members` vec".

### `Execute`

Expects accounts:
- `accounts[0]`: `multisig_state`
- `accounts[1]`: executor
- `accounts[2]`: `proposal` PDA
- `accounts[3..]`: `N` target accounts (variable, determined by `proposal.target_account_count`)

The number of target accounts is only known at runtime (stored in the proposal). The framework currently cannot express this.

### Option A: Trailing `Vec<AccountWithMetadata>` parameter

```rust
#[instruction]
pub fn create_multisig(
    #[account(init, pda = ...)]
    multisig_state: AccountWithMetadata,
    create_key: [u8; 32],
    threshold: u8,
    members: Vec<[u8; 32]>,
    // Trailing accounts — count must match `members.len()`
    #[account(repeat = "members")]
    member_accounts: Vec<AccountWithMetadata>,
) -> NssaResult { ... }
```

### Option B: Separate `#[remaining_accounts]` annotation

```rust
#[instruction]
pub fn execute(
    #[account(...)]
    multisig_state: AccountWithMetadata,
    #[account(signer)]
    executor: AccountWithMetadata,
    #[account(...)]
    proposal: AccountWithMetadata,
    proposal_index: u64,
    #[remaining_accounts]
    target_accounts: Vec<AccountWithMetadata>,
) -> NssaResult { ... }
```

## Impact on macro-generated dispatch

The `main()` generated by `#[nssa_program]` currently does:
```rust
let [acc1, acc2, acc3] = <[_; 3]>::try_from(pre_states).unwrap();
```

With variable accounts, it would need:
```rust
let (fixed_accounts, remaining) = pre_states.split_at(3);
let [acc1, acc2, acc3] = fixed_accounts.try_into().unwrap();
```

## Blocking

[...]

PR #209, opened 2026-05-18
feat(spel-client-gen): --target logos-module — Qt/QML Basecamp module scaffold from IDL

### LINKED ISSUE #207: feat(spel-client-gen): --target logos-module — generate C++ Qt/QML plugin scaffold from IDL for Basecamp UI modules
## Summary

`spel-client-gen` currently generates a Rust client and C FFI wrappers from the SPEL IDL. When building a **Basecamp UI module** (the QML-based Logos application platform), there is a large manual gap between the generated C header and a working, loadable module: every developer must hand-write 200–400 lines of identical C++ Qt boilerplate.

This issue proposes adding a `--target logos-module` flag to `spel-client-gen` that closes that gap by emitting a complete, compilable C++ QML plugin scaffold directly from the IDL.

---

## Current Workflow and Where It Breaks

The end-to-end pipeline for turning a SPEL program into a usable Basecamp app today looks like this:

```
Rust #[lez_program] annotations
        │
        ▼  (make generate-idl)
multisig_idl.json                   ← SpelIdl struct, fully machine-readable
        │
        ▼  (spel-client-gen --idl ... --out-dir ...)
{name}_ffi.rs  +  {name}.h          ← spel-client-gen stops here
        │
        ▼  ✋ MANUAL — currently hand-written per project
XyzBackend.h / XyzBackend.cpp       ← QObject with Q_INVOKABLE per instruction
XyzPlugin.h  / XyzPlugin.cpp        ← QML plugin registration
module.yaml                         ← logos-module-builder config
Main.qml                            ← QML UI scaffold
        │
        ▼  (logos-module-builder / nix build)
libxyz_module.so + metadata.json    ← loadable Basecamp module
```

Everything from `XyzBackend.*` downward is mechanical, repetitive, and purely derived from the IDL — yet every module author must write it from scratch. The [lez-multisig Basecamp module](https://github.com/logos-co/lez-multisig) we are currently building exposed this gap clearly.

### What "mechanical" means here

For every `#[instruction]` in the IDL, the C++ backend needs:

```cpp
// For each instruction — ~15 lines each, identical structure every time
Q_INVOKABLE void approve(quint64 proposalIndex) {
    if (m_busy) return;
    setBusy(true);
    auto args = QJsonObject{{"proposal_index", (qint64)proposalIndex},
                            {"create_key",     m_createKey},
                            {"program_id_hex", m_programIdHex},
                            {"wallet_path",    m_walletPath}};
    QThreadPool::globalInstance()->start([this, args]() {
        auto json = QJsonDocument(args).toJson(QJsonDocument::Compact);
        auto raw  = lez_multisig_approve(json.constData());
        auto res  = QString::fromUtf8(raw);
        lez_multisig_free_string(raw);
        QMetaObject::invokeMethod(this, [this, res]() {
            setBusy(false);
            auto obj = QJsonDocument::fromJson(res.toUtf8()).object();
            if (obj["success"].toBool()) emit operationSuccess("approve", obj["tx_hash"].toString());
            else                         emit operationError  ("approve", obj["error"].toString());
        }, Qt::QueuedConnection);
    });
}
```

Plus plugin registration (`Q_PLUGIN_METADATA`, `qmlRegisterSingletonType`), CMake wiring, and a `module.yaml` for [logos-module-builder](https://github.com/logos-co/logos-workspace/tree/main/repos/logos-module-builder).

---

## Proposed Solution

Add a `--target logos-module` flag (or subcommand) to `spel-client-gen`:

```bash
spel-client-gen \
  --idl multisig_idl.json \
  --out-dir lez-multisig-module \
  --target logos-module \
  --module-name lez_multisig \
  --ffi-lib lez_multisig_ffi
```

#### 1. `{name}_backend.h` / `{name}_backend.cpp`

A `QObject` subclass exposing:

- **One `Q_INVOKABLE` per instruction** — each runs the FFI call on a `QThreadPool` worker and emits a result signal on the Qt thread. Argument names and types come directly from `IdlInstruction.args` and `IdlArg.type_`.
- **`Q_PROPERTY busy`** — true while any FFI call is in flight.
- **`Q_PROPERTY lastError`** — last error string, cleared on next call start.
- **`Q_PROPERTY walletPath` / `programIdHex`** — persisted via `QSettings`.
- **`operationSuccess(QString op, QString txHash)`** and **`operationError(QString op, QString error)`** signals — the standard Basecamp result pattern (see [whisper-wall](https://github.com/logos-co/whisper-wall/blob/main/ui/qml/Main.qml#L45-L55)).
- A `refreshState()` slot that calls any read-only query functions (IDL instructions with no signer/writable accounts).

Type mapping from IDL to Qt (`IdlType` → `QMetaType`):
| IDL type | C++ parameter type |
|---|---|
| `u8`, `u16`, `u32` | `quint32` |
| `u64` | `quint64` |
| `i64` | `qint64` |
| `bool` | `bool` |
| `string` | `QString` |
| `bytes` / `[u8;32]` | `QString` (hex) |
| `pubkey` | `QString` (hex) |
| `vec<T>` | `QStringList` or `QVariantList` |

#### 2. `{name}_plugin.h` / `{name}_plugin.cpp`

Standard Qt QML plugin boilerplate:

```cpp
class LezMultisigPlugin : public QQmlExtensionPlugin {
    Q_OBJECT
    Q_PLUGIN_METADATA(IID QQmlExtensionInterface_iid)
public:
    void registerTypes(const char *uri) override {
        qmlRegisterSingletonType<LezMultisigBackend>(
            uri, 1, 0, "LezMultisigBackend",
            [](QQmlEngine*, QJSEngine*) -> QObject* {
                return new LezMultisigBackend();
            });
    }
};
```

This is 100% identical across all modules — only class names change.

#### 3. `module.yaml`

Ready for [logos-module-builder](https://github.com/logos-co/logos-workspace/tree/main/repos/logos-module-builder) consumption:

```yaml
name: lez_multisig
version: 0.1.0        # from SpelIdl.version
type: ui
category: tools
description: "Multisig governance UI for LEZ"

nix:
  packages:
    build: [qtbase, cmake]
    runtime: [qt6.qtdeclarative]
  external_libraries:
    - name: lez_multisig_ffi
      vendor_path: lib

cmake:
  find_packages: [Qt6Core, Qt6Quick, Qt6Qml]
  link_libraries: [Qt6::Core, Qt6::Quick, Qt6::Qml, lez_multisig_ffi]
  extra_sources:
    - src/lez_multisig_backend.cpp
    - src/lez_multisig_plugin.cpp
```

#### 4. `Main.qml` stub

A minimal but runnable QML file with one section per instruction — not production UI, but something that compiles, loads in Basecamp, and calls the backend correctly on day one:

```qml
import QtQuick
import QtQuick.Controls
import QtQuick.Layouts

Rectangle {
    id: root
    color: "#0f1117"
    // ... palette ...

    // Generated from IDL: one collapsible section per instruction
    // approve: { proposal_index: u64 }
    // reject:  { proposal_index: u64 }
    // execute: { proposal_index: u64 }
    // ...

    Connections {
        target: backend
        function onOperationSuccess(op, txHash) { toast.show("✓ " + op, true)  }
        function onOperationError(op, err)       { toast.show("✗ " + err, false) }
    }
}
```

---

## Relationship to logos-module-builder

[logos-module-builder](https://github.com/logos-co/logos-workspace/tree/main/repos/logos-module-builder) already handles the **build layer** (Nix derivations, CMake wiring, packaging). It has `templates/ui-module/` and `templates/ui-qml-module/` that describe the expected file layout. The generator proposed here would produce exactly those expected files, making the two tools complementary:

```
spel-client-gen --target logos-module   →  generates source scaffold
logos-module-builder (nix build)        →  compiles + packages the module
```

No changes to logos-module-builder are needed; this proposal lives entirely within `spel-client-gen`.

---

## Scope and Non-Goals

**In scope:**
- C++ backend class (`Q_INVOKABLE` per instruction + standard signals/properties)
- QML plugin registration boilerplate
- `module.yaml` for logos-module-builder
- Stub `Main.qml` with one form section per instruction

**Out of scope (intentionally):**
- Production UI design — the generated `Main.qml` is a starting point, not a finished app
- Read-only query helpers (e.g. `list_proposals`, `get_state` in lez-multisig) that are not in the IDL — these continue to be hand-written and added to the backend manually
- Localization, accessibility, theming beyond basic palette

---

## Motivation / Prior Art

- `spel-client-gen` already reads `SpelIdl` and generates typed code — the infrastructure for IDL-driven generation exists and is well-structured ([`spel-client-gen/src/lib.rs`](https://github.com/logos-co/spel/blob/main/spel-client-gen/src/lib.rs), [`ffi_codegen.rs`](https://github.com/logos-co/spel/blob/main/spel-client-gen/src/ffi_codegen.rs))
- [Anchor](https://www.anchor-lang.com/) (Solana) generates client SDKs from IDL; this is the Qt/C++ equivalent for Logos
- whisper-wall and joke-wall (Basecamp reference apps) both have identical plugin registration boilerplate — clear signal that it should be generated, not copied
- lez-multisig is the immediate consumer: building its Basecamp UI module requires writing all this by hand today

---

## Suggested Implementation Path

1. Add `--target` flag to `spel-client-gen` CLI (`main.rs`) — default `rust-ffi` (current behaviour, no regression)
2. Add `qt_codegen.rs` module in `spel-client-gen/src/` with `generate_qt_backend()` and `generate_qt_plugin()`
3. Add `module_yaml.rs` for `generate_module_yaml()`
4. Add `qml_stub.rs` for `generate_main_qml()`
5. Extend `CodegenOutput` struct with optional Qt fields, or use a separate `QtCodegenOutput`
6. Add integration test: run against `lez-multisig` IDL, assert generated files compile with Qt6

The Qt codegen module follows the same pattern as the existing `ffi_codegen.rs` — string-building against the `SpelIdl` struct — so the implementation cost is moderate and well-contained.

vpavlin:
## Summary

Implements the code generation target proposed in #207. Adds `--target logos-module` to `spel-client-gen` which emits a complete, compilable Qt/QML Basecamp module scaffold directly from a SPEL IDL.

### Generated output (7 files)

```
spel-client-gen --idl my_program_idl.json --out-dir out/ --target logos-module

out/
  src/MyProgramBackend.h      QObject: Q_INVOKABLE per instruction, Q_PROPERTY per state account
  src/MyProgramBackend.cpp    Async FFI dispatch (QFutureWatcher + QThreadPool), baseArgs, handleFfiResult
  src/MyProgramPlugin.h       QML singleton plugin header
  src/MyProgramPlugin.cpp     qmlRegisterSingletonType boilerplate
  qml/Main.qml                Functional UI: inputs+button per instruction, key/value display per state
  module.yaml                 logos-module-builder build config (external_libraries, cmake sources)
  metadata.json               Basecamp runtime metadata (main, view, nix packages)
```

### Account → Qt mapping

| IDL account | Generated |
|---|---|
| `is_signer && !pda` | `const QString&` param on Q_INVOKABLE |
| PDA (has `seeds`) | No param — computed internally by FFI |
| State account (`idl.accounts` struct with fields) | `QVariantMap` Q_PROPERTY + `fetch{Name}()` method |

### Tested against

- **whisper-wall IDL** — 5 instructions, no state accounts → 5 Q_INVOKABLE methods, full QML with input fields
- **lez-multisig FFI IDL** — 8 instructions, no `accounts` types → 8 Q_INVOKABLE methods; comparison with hand-written `LezMultisigBackend` confirms the scaffold gives the right base for layering domain-specific state management on top

All 39 existing tests pass.

## Test plan

- [ ] `cargo test -p spel-client-gen` — all 39 pass
- [ ] `spel-client-gen --idl whisper-wall-idl.json --out-dir out/ --target logos-module` — generates 7 files
- [ ] `spel-client-gen --idl lez-multisig-ffi/src/multisig_idl.json --out-dir out/ --target logos-module` — generates 7 files, Q_INVOKABLE per instruction
- [ ] Default behaviour unchanged: `spel-client-gen --idl ... --out-dir ...` still generates Rust client + FFI + header

Closes #207


vpavlin (commented):
## Review: tested against `lez-multisig` IDL (real-world evaluation)

I ran `spel-client-gen --target logos-module` against the lez-multisig IDL and tried to drop the output into the [lez-multisig](https://github.com/logos-co/lez-multisig) project. Branch: `feat/spel-logos-module-codegen`. Here are the issues found, ordered by severity.

---

### 🔴 Bugs (prevent building or silently fail at runtime)

**1. `QString(QVariant)` is not valid in Qt6 — compile error**

Generated pattern for all `QVariantList` parameters:
```cpp
for (const auto& _s : members) _arr.append(QString(_s));
```
Qt6 removed the implicit `QVariant → QString` conversion. This causes a hard compile error. Fix: `_s.toString()`.
Affects every instruction that takes a `Vec<String>` / array parameter.

**2. Env var name doubles the program name — silent runtime failure**

Generated constructor:
```cpp
m_programIdHex(qEnvironmentVariable("MULTISIG_PROGRAM_PROGRAM_ID_HEX"))
```
The IDL program name `multisig_program` is doubled, producing `MULTISIG_PROGRAM_PROGRAM_ID_HEX`. Should be `MULTISIG_PROGRAM_ID_HEX` (or the user's canonical env var — but definitely not doubled).

**3. Wrong plugin interface for Basecamp**

Generator emits:
```cpp
class MultisigProgramPlugin : public QQmlExtensionPlugin {
    Q_OBJECT
    Q_PLUGIN_METADATA(IID QQmlExtensionInterface_iid)
public:
    void registerTypes(const char* uri) override;
};
```
Basecamp modules use a `createWidget(LogosAPI*) → QWidget*` / `destroyWidget(QWidget*)` interface (custom `IComponent`). `QQmlExtensionPlugin` is for standalone Qt QML modules, not Basecamp plugins. Needed a full manual rewrite of both Plugin files.

**4. QML uses `backend` but Plugin registers `Backend` (capital B)**

Generated QML: `backend.busy`, `backend.createMultisig(...)` etc. (lowercase `backend` context property).
Generated Plugin: `qmlRegisterSingletonType<...>(uri, 1, 0, "Backend", ...)` (capital `B`, QML singleton).
These don't match. QML would silently get `undefined` for every property and method call.

---

### 🟡 Missing features (functional gaps vs. a production-ready module)

**5. Array parameters passed as raw strings in QML**

```qml
onClicked: backend.createMultisig(..., create_multisig_members_f.text)
```
The C++ method takes `QVariantList` but a plain `TextField.text` (string) is passed. Qt/QML silently wraps it in a single-element list. For `members` (which should be a list of public keys), the user needs a newline/comma-split. Should generate something like:
```qml
create_multisig_members_f.text.split("\n").map(s => s.trim()).filter(s => s.length > 0)
```

**6. No state properties — backend is instruction-dispatch only**

The generated backend has no `currentMultisig`, `proposals`, `members`, or `multisigs` Q_PROPERTYs. There's no way to display multisig state in QML after an operation succeeds. A minimal addition would be a generic `lastResult` (QVariantMap) property populated from the FFI response.

**7. No refresh-on-success pattern**

`handleFfiResult()` emits `operationSuccess` but never re-fetches state. For read-heavy UIs a `QTimer::singleShot(1200, ...)` to re-query would be a helpful default.

**8. Instruction methods expose raw signer parameters**

```cpp
Q_INVOKABLE void approve(const QString& approverId, quint64 proposalIndex, const QString& createKey);
```
The wallet already knows the signing key. Requiring users to type their own public key every time is poor UX. For signer accounts, the generated method could either omit the parameter (wallet signs implicitly) or at minimum document why it's exposed.

**9. No `QSettings` persistence**

Wallet path and program ID are read once from env vars and lost on restart. A single `QSettings("org", "app")` in the constructor would make dev much smoother.

---

### 🔵 Naming / UX

**10. Generated module name uses raw IDL name verbatim**

Output: `multisig_program`, `libmultisig_program_plugin`, `multisig_program_ffi`.
Project's canonical name: `lez_multisig` / `liblez_multisig_plugin` / `lez_multisig_ffi`.
A `--name <name>` flag (or stripping a `_program` suffix by convention) would help.

**11. No `<prog>_program_id()` FFI call in the constructor**

The compile-time program ID embedding pattern (baked in via `build.rs` + `cargo:rustc-env`, exposed as `lez_multisig_program_id()`) is not reflected in the generated constructor. The generator can't know about this without IDL support, but it could at least document the env var pattern or leave a `// TODO: call <prog>_program_id() here if baked-in` placeholder.

---

Overall the structural bones are good — async dispatch, JSON marshalling, palette, toast notifications all work. The blocking issues are #1 (Qt6 compile error), #3 (wrong plugin interface), and #4 (QML/plugin name mismatch). Happy to provide patches for any of these.

vpavlin:
Thanks for the thorough review — all 11 points addressed in the latest commit (f21d471).

**Blocking bugs fixed:**

1. **Qt6 compile error on `QString(_s)`** — split `arg_to_json_lines` by inner type: `QStringList` iterates `const QString&` directly, `QVariantList` uses `QJsonValue::fromVariant`. No implicit `QVariant→QString`.

2. **Doubled env-var suffix** — strip `_program`/`_contract` before uppercasing: `multisig_program` → `MULTISIG` → `MULTISIG_PROGRAM_ID_HEX` (not `MULTISIG_PROGRAM_PROGRAM_ID_HEX`).

3. **Wrong plugin interface** — switched from `QQmlExtensionPlugin::registerTypes()` to the correct Basecamp `IComponent` interface with `createWidget(LogosAPI*)` / `destroyWidget(QWidget*)`, `Q_PLUGIN_METADATA(IID IComponent_iid FILE "../metadata.json")`.

4. **QML `backend` vs `Backend` mismatch** — resolved by fix #3: plugin now calls `setContextProperty("backend", m_backend)` so QML's lowercase `backend` is correct.

**Other issues addressed:**

5. Vec params in QML now use a `TextArea` with the split expression: `.text.split("\n").map(s => s.trim()).filter(s => s.length > 0)`.
6. Added `Q_PROPERTY(QVariantMap lastResult ...)` populated from `obj.toVariantMap()` on every successful FFI call.
7. Added `autoRefresh()` private Q_SLOT + `QTimer::singleShot(1200, ...)` after successful TX to refresh all no-arg fetch methods.
8. Added `src/main.cpp` standalone preview app (was missing from the file list).
9. Added `--module-name <name>` CLI flag so class/file/env-var names can be overridden independently of the IDL `name` field (e.g. `--module-name lez_multisig` gives `LezMultisigBackend` from a `multisig_program` IDL).
10–11. `module.yaml` and `metadata.json` also use the effective module name.

Tested end-to-end with the lez-multisig IDL:
```
spel-client-gen --idl multisig.json --out-dir out/ --target logos-module --module-name lez_multisig
```
Produces `LezMultisigBackend.h/.cpp`, `LezMultisigPlugin.h/.cpp`, `src/main.cpp`, `qml/Main.qml`, `module.yaml`, `metadata.json` — all compiling cleanly (verified locally).

vpavlin (commented):
## Follow-up: re-tested after fix commit (f21d471)

Re-ran `spel-client-gen --target logos-module --module-name lez_multisig` against the same lez-multisig IDL. Branch updated: `feat/spel-logos-module-codegen`. Builds clean.

### ✅ Fixed

| # | Issue | How fixed |
|---|-------|-----------|
| 1 | `QString(QVariant)` Qt6 compile error | Changed to `QJsonValue::fromVariant(_v)` — correct and idiomatic |
| 2 | Env var name doubling (`MULTISIG_PROGRAM_PROGRAM_ID_HEX`) | Fixed; with `--module-name lez_multisig` produces `LEZ_MULTISIG_PROGRAM_ID_HEX` |
| 3 | Wrong plugin interface (`QQmlExtensionPlugin`) | Now generates proper `IComponent` with `createWidget`/`destroyWidget` |
| 4 | QML `backend` vs Plugin `Backend` mismatch | Plugin now uses `setContextProperty("backend", m_backend)` |
| 5 | Array params passed as raw strings in QML | Now generates `.split("\n").map(...).filter(...)` correctly |
| 10 | No way to override generated module name | `--module-name <name>` flag added — works for class names, filenames, and top-level metadata |

Also: `main.cpp` is now generated too — nice addition.

### ⚠️ Partial fix: `--module-name` doesn't update the FFI library reference

With `--module-name lez_multisig`, the external library in `module.yaml` and `metadata.json` still uses the IDL name:
```yaml
# module.yaml — still wrong:
external_libraries:
  - name: multisig_program_ffi   # should be lez_multisig_ffi
```
```json
// metadata.json — still wrong:
"external_libraries": [{"name": "multisig_program_ffi"}]
```
The assumption is `<module-name>_ffi` but the override isn't applied there. Small fix — the FFI lib name should follow `--module-name`.

### Still open (not addressed by the fix commit)

| # | Issue |
|---|-------|
| 6 | No state properties (`currentMultisig`, `proposals`, etc.) |
| 7 | No refresh-on-success after transactions |
| 8 | Signer accounts exposed as method parameters (user must type their own pubkey) |
| 9 | No `QSettings` persistence for wallet path / program ID |
| 11 | No `<module>_program_id()` compile-time program ID call in constructor |

These are design-level gaps rather than bugs — the generated scaffold is a working starting point, just not production-complete. Happy to discuss which of these the generator should handle vs. leaving to the developer.

vpavlin:
Thanks for re-testing. Follow-up addressed in `dddd21a`:

**Fixed:**
- `--module-name` now propagates to `external_libraries.name` in both `module.yaml` and `metadata.json`. `--module-name lez_multisig` now correctly produces `lez_multisig_ffi` in both files.

**Open items (#6, #7, #8, #9, #11):**

These are design-level decisions — happy to implement any you want, but wanted to check first:

- **#6 (state properties)**: The IDL doesn't distinguish "state accounts" from "instruction accounts" in a way the generator can reliably detect. `lastResult: QVariantMap` is the generic fallback — would a smarter heuristic help, or is this left to the developer?
- **#7 (refresh-on-success)**: `autoRefresh()` + `QTimer::singleShot(1200, ...)` is already generated in this commit — it calls all no-arg fetch methods 1.2s after a successful TX. If you're still seeing it missing, let me know what you expected.
- **#8 (signer params)**: The generator passes them through because the wallet path is caller-supplied via env — removing them would require the backend to know which key to use. A `// Signer: wallet provides this key` comment in the generated method might be enough?
- **#9 (QSettings)**: Can add `QSettings` for wallet path + program ID hex. Straightforward — say the word.
- **#11 (compile-time program ID)**: This requires a `<prog>_program_id()` FFI symbol that may or may not exist. Could add a `// TODO: replace with <prog>_program_id() if available` comment in the constructor.

vpavlin (commented):
## Follow-up 2: re-tested after Copilot review fixes (dddd21a9)

Re-ran `spel-client-gen --target logos-module --module-name lez_multisig`. Builds clean. Branch updated: `feat/spel-logos-module-codegen`.

### ✅ All build-blocking issues resolved

The remaining partial fix from my last review — `--module-name` not propagating to the FFI library name in `module.yaml`/`metadata.json` — is now fixed. With `--module-name lez_multisig` the output correctly produces `lez_multisig_ffi` throughout.

### 👍 Bonus: `lastResult` Q_PROPERTY added

`handleFfiResult()` now populates a `lastResult` QVariantMap from every successful FFI response and emits `lastResultChanged()`. This is a partial address of issue #6 (no state properties) — QML can now react to operation results without polling. Good addition.

Also noticed `QTimer` is now included in the header, suggesting refresh support may be coming.

### Still open (design gaps, not bugs)

| # | Issue |
|---|-------|
| 6 | `lastResult` helps but there's no persistent state (`currentMultisig`, `proposals`) — each call's response is ephemeral |
| 7 | No refresh-on-success (though `QTimer` include hints this may land) |
| 8 | Signer accounts still exposed as method parameters |
| 9 | No `QSettings` persistence |
| 11 | No `<module>_program_id()` compile-time program ID call (workaround: env var documented via comment at bottom of header — nice touch) |

At this point the generator output is solid as a starting scaffold. The remaining gaps are expected for a v1 — a developer can build on top of what's generated. LGTM for merging.

vpavlin:
## Follow-up testing report (lez-multisig integration)

Rebuilt and reinstalled the plugin into a live Basecamp instance using the latest `feat/207-logos-module-codegen` tip. The generation is clean — no manual fixes needed for the previously reported issues (Qt6 string conversion, env var name, plugin interface, context property name all correct now). Two remaining issues found:

---

### Bug 1: `qrc:/qml/Main.qml` silently fails when loaded via `QPluginLoader`

**File**: `src/LezMultisigPlugin.cpp`, `createWidget()`

When Basecamp loads the plugin via `QPluginLoader::load()`, the Qt resource system does not automatically register the embedded QML resource. The `qrc:/qml/Main.qml` URL is therefore unresolvable and the widget renders as an empty white rectangle with no error shown.

The fix is to call `Q_INIT_RESOURCE` explicitly before `setSource`:

```cpp
} else {
    Q_INIT_RESOURCE(lez_multisig_qml);  // must be called explicitly in dynamically loaded plugins
    view->setSource(QUrl("qrc:/qml/Main.qml"));
}
```

Note: the resource name (`lez_multisig_qml`) must match the name passed to `qt_add_resources()` in `CMakeLists.txt`. The generator should emit this call, using the same name it generates for the cmake resource target.

With this fix the embedded resource loads correctly without needing `QML_PATH`.

---

### Potential issue 2: `ColumnLayout` padding properties

**File**: `qml/Main.qml`

The generated QML uses `leftPadding`, `rightPadding`, `topPadding`, `bottomPadding` on `ColumnLayout`:

```qml
ColumnLayout {
    width: scrollView.width
    spacing: 12
    leftPadding: 16; rightPadding: 16
    topPadding: 16; bottomPadding: 80
```

These properties were added to `QtQuick.Layouts` in Qt 6.6, so they work with Qt 6.9.2. However on older Qt 6.x builds (< 6.6) this will produce a runtime error: `Cannot assign to non-existent property "bottomPadding"`. Worth noting in docs/changelog if the target environment might have Qt < 6.6.

---

Both issues found on `lez-multisig` @ `feat/spel-logos-module-codegen`.

vpavlin:
## Design gap: configuration properties not exposed or persisted

Testing the generated scaffold against a real Basecamp deployment surfaced a UX gap that the generator should address. Three values that every LEZ module needs are currently only readable from environment variables at construction time, with no way to see, override, or persist them from the UI:

| Value | Env var read in constructor | Q_PROPERTY | Setter | QSettings |
|---|---|---|---|---|
| `wallet_path` | `NSSA_WALLET_HOME_DIR` | ❌ | ❌ | ❌ |
| `sequencer_url` | `NSSA_SEQUENCER_URL` | ❌ | ❌ | ❌ |
| `program_id_hex` | `LEZ_MULTISIG_PROGRAM_ID_HEX` | ❌ | ❌ | ❌ |

### What users actually need

1. **See the current values** — a settings panel in the QML showing which wallet, sequencer, and program they are connected to.
2. **Override from the UI** — text fields that let the user change any of the three without restarting the app or setting env vars.
3. **Persist across restarts** — overrides should survive app restart (QSettings with a per-program-name group is the natural fit in Qt).

Priority chain should be: `QSettings` > env var > compiled-in default, so that a user-entered value sticks but env vars can still override for scripted/CI use.

### Suggested generated backend additions

```cpp
// In LezMultisigBackend.h
Q_PROPERTY(QString walletPath   READ walletPath   WRITE setWalletPath   NOTIFY walletPathChanged)
Q_PROPERTY(QString sequencerUrl READ sequencerUrl WRITE setSequencerUrl NOTIFY sequencerUrlChanged)
Q_PROPERTY(QString programIdHex READ programIdHex WRITE setProgramIdHex NOTIFY programIdHexChanged)

Q_INVOKABLE void setWalletPath(const QString& v);
Q_INVOKABLE void setSequencerUrl(const QString& v);
Q_INVOKABLE void setProgramIdHex(const QString& v);
```

```cpp
// In LezMultisigBackend.cpp constructor — priority: QSettings > env var > default
QSettings s("logos-co", "<module_name>");
m_walletPath   = s.value("walletPath",   qEnvironmentVariable("NSSA_WALLET_HOME_DIR", ".scaffold/wallet")).toString();
m_sequencerUrl = s.value("sequencerUrl", qEnvironmentVariable("NSSA_SEQUENCER_URL",   "http://127.0.0.1:3040")).toString();
m_programIdHex = s.value("programIdHex", qEnvironmentVariable("LEZ_MULTISIG_PROGRAM_ID_HEX")).toString();
```

```cpp
void LezMultisigBackend::setWalletPath(const QString& v) {
    if (m_walletPath == v) return;
    m_walletPath = v;
    QSettings("logos-co", "<module_name>").setValue("walletPath", v);
    emit walletPathChanged();
}
// (same pattern for sequencerUrl, programIdHex)
```

### Suggested QML additions

A small collapsible settings section at the top or bottom of the generated QML — three labelled text fields bound to `backend.walletPath`, `backend.sequencerUrl`, `backend.programIdHex` with `onEditingFinished` calling the setter. This makes the module self-contained and usable without any env var ceremony.

This is a systematic gap rather than a per-program quirk — every `--target logos-module` scaffold will have the same issue.

vpavlin:
## Follow-up 3: sidebar QML redesign + QSettings + Q_INIT_RESOURCE (265b8e4)

Re-ran against the same lez-multisig IDL. All previously open blocking bugs remain fixed. New changes:

### ✅ Fixed in 265b8e4

**Bug from integration testing (reported above):**
- **`Q_INIT_RESOURCE` in `createWidget()`** — Qt does not auto-register embedded QRC resources in dynamically loaded plugins. `Q_INIT_RESOURCE(lez_multisig_qml)` is now called explicitly before `setSource(QUrl("qrc:/qml/Main.qml"))`. The resource name is derived from the effective module name (with `-` → `_`), matching what `qt_add_resources()` in `CMakeLists.txt` should use.

**Design gap #9 (QSettings persistence):**
- All three config values (`walletPath`, `sequencerUrl`, `programIdHex`) now have `Q_PROPERTY` with getter/setter/signal triad.
- Constructor uses priority chain: `QSettings > qEnvironmentVariable(envVar, default)` — persisted values survive restart, env vars still work for scripted/CI use, hardcoded defaults as last resort.
- Setters persist immediately via `QSettings("logos-co", "<module_name>").setValue(key, v)`.

**QML redesign:**
- Replaced the flat single-page QML with a **sidebar + `StackLayout`** layout:
  - Left: 200px `colSidebar` panel with `ListView` nav (program title + BusyIndicator header, then one item per fetch account, one per instruction, then Settings)
  - Right: `StackLayout` with one `Item > ScrollView > ColumnLayout` page per nav item
  - Fetch pages show a refresh button + seed arg fields + `Repeater` over `backend.<account>` keys
  - Instruction pages unchanged (fields + submit button)
  - **Settings page** (always last): three labelled TextFields bound to `backend.walletPath`, `backend.sequencerUrl`, `backend.programIdHex` with `onEditingFinished` calling setters

**Note on `ColumnLayout` padding (reported above):**
The new QML uses `Layout.leftMargin`/`Layout.rightMargin` on individual items instead of `leftPadding`/`rightPadding` on `ColumnLayout`, so there's no Qt < 6.6 issue.

### Still open (design gaps)

| # | Issue |
|---|-------|
| 6 | No persistent named-state properties (`currentMultisig`, `proposals`) — `lastResult` is the generic fallback |
| 7 | No auto-refresh after TX (the `QTimer` include is there; left to the developer for now) |
| 8 | Signer accounts still exposed as method parameters |
| 11 | No `<module>_program_id()` compile-time program ID call |

These remain developer-supplied customizations on top of the scaffold as before.

vpavlin:
## Suggestion: standardise `{module_name}_program_id()` so the generator can emit it

### Context

The generated constructor currently resolves `programIdHex` with a two-level priority chain:

```
QSettings → env var (LEZ_MULTISIG_PROGRAM_ID_HEX) → ""
```

In practice the env var is only set in CI or scripted runs. A deployed module running inside Basecamp has neither, so `programIdHex` starts blank and every transaction fails — the user has to paste the hex into the settings panel before doing anything.

The real value is already available at FFI compile time: it is the RISC0 image ID baked in by `build.rs`. We added a thin FFI export to surface it:

```rust
// lez-multisig-ffi/src/lib.rs
#[no_mangle]
pub extern "C" fn lez_multisig_program_id() -> *mut c_char {
    to_cstring(format!(r#"{{"program_id_hex":"{}"}}"#, env!("MULTISIG_PROGRAM_ID_HEX", "")))
}
```

This is a manual patch today. It should be a generator-emitted standard.

### Proposed convention

Every `--target logos-module` FFI crate exposes a no-arg function named `{module_name}_program_id()` that returns the compiled-in program ID as JSON `{"program_id_hex":"<64-char hex>"}` (or an empty string if not baked in). The SPEL scaffold/framework can provide a macro to make this trivial:

```rust
// spel_export_program_id!("MULTISIG_PROGRAM_ID_HEX");
// expands to:
#[no_mangle]
pub extern "C" fn lez_multisig_program_id() -> *mut c_char { ... }
```

### Generator changes needed

**`extern "C"` block** — add one line alongside the existing instruction declarations:

```cpp
extern "C" {
    char* multisig_program_create_multisig(const char* args_json);
    // ... other instructions ...
    char* lez_multisig_program_id();          // ← add: {effective_prog}_program_id
    void  multisig_program_free_string(char* s);
}
```

**Constructor** — extend to use it as a third fallback:

```cpp
m_programIdHex = s.value("programIdHex", qEnvironmentVariable("LEZ_MULTISIG_PROGRAM_ID_HEX")).toString();
if (m_programIdHex.isEmpty()) {
    char* raw = lez_multisig_program_id();
    if (raw) {
        m_programIdHex = QJsonDocument::fromJson(QByteArray(raw))
                             .object().value("program_id_hex").toString();
        multisig_program_free_string(raw);
    }
}
```

Full priority chain becomes: **QSettings → env var → compiled-in FFI constant → ""**

### Naming rationale

- Function name uses `{effective_prog}` (the `--module-name` value, e.g. `lez_multisig`) not the raw IDL name (`multisig_program`) — it is a module-identity concern, not a program-instruction concern.
- Freed with the existing `{prog}_free_string` since both allocate via the same Rust allocator.
- No IDL change required — the generator derives the name from `--module-name`.

This makes a deployed module work out of the box without env vars or manual settings.

vpavlin:
## Follow-up 4: `{effective_prog}_program_id()` FFI fallback (71a8902)

Implemented as suggested. The full priority chain is now:

**QSettings → `qEnvironmentVariable` → `{effective_prog}_program_id()` FFI call → `""`**

### Changes

**`extern "C"` block** — one new declaration alongside instructions and fetch functions:
```cpp
extern "C" {
    char* lez_multisig_create(const char* args_json);
    // ... instructions, fetch functions ...
    char* lez_multisig_program_id();   // ← new
    void  lez_multisig_free_string(char* s);
}
```

**Constructor** — fallback block after the QSettings/env-var line:
```cpp
m_programIdHex = s.value("programIdHex", qEnvironmentVariable("LEZ_MULTISIG_PROGRAM_ID_HEX")).toString();
if (m_programIdHex.isEmpty()) {
    char* raw = lez_multisig_program_id();
    if (raw) {
        m_programIdHex = QJsonDocument::fromJson(QByteArray(raw))
                             .object().value("program_id_hex").toString();
        lez_multisig_free_string(raw);
    }
}
```

The function name uses `effective_prog` (the `--module-name` value), freed with `{prog}_free_string` as suggested. The FFI crate just needs to export `lez_multisig_program_id()` returning `{"program_id_hex":"<hex>"}` — no IDL or generator-side changes needed on the Rust side beyond what you already added manually.

vpavlin:
## Feedback: `[u8; 32]` instruction args don't accept base58 account IDs

### Problem

Any instruction arg typed as `{"array": ["u8", 32]}` (or `{"vec": {"array": ["u8", 32]}}`) is currently parsed with raw `serde_json::from_value`, which only accepts a JSON byte array. In practice, users enter member account IDs from their wallet as **base58 strings** (with or without a `Public/`/`Private/` prefix), not JSON byte arrays.

The affected fields in the multisig FFI:
- `members` in `create_multisig` (`Vec<[u8; 32]>`)
- `new_member` in `propose_add_member` (`[u8; 32]`)
- `member` in `propose_remove_member` (`[u8; 32]`)

The error at runtime: `parse error: invalid type: string "Public/CbgR6tj5kWx5oziiFptM7jMvrQeYY3Mzaao6ciuhSr2r", expected a sequence`.

Note: the `create_key` field was already patched locally (in `lez-multisig`) with a `parse_create_key()` helper that accepts a hex string or JSON array, but that doesn't cover base58 (since `create_key` is a random nonce, not an account ID).

### Reference: how the SPEL CLI handles this

`spel-cli/src/hex.rs` defines `decode_bytes_32`:
```rust
pub fn decode_bytes_32(input: &str) -> Result<[u8; 32], String> {
    let input = input
        .strip_prefix("Public/")
        .or_else(|| input.strip_prefix("Private/"))
        .unwrap_or(input);

    if let Ok(bytes) = input.from_base58() {
        if bytes.len() == 32 { ... return Ok(arr); }
    }

    let hex = input.strip_prefix("0x").or_else(|| input.strip_prefix("0X")).unwrap_or(input);
    let bytes = hex_decode(hex)?;
    ...
}
```

This is what `parse_account_id` in the generated FFI is based on (for account-slot strings), but it isn't applied to instruction args of the same type.

### Suggested fix for the generator

Generate a single `parse_bytes32` helper that accepts **base58 (with/without `Public/`/`Private/` prefix), hex (with/without `0x`), or JSON byte array**, and use it for every `{"array": ["u8", 32]}` instruction arg instead of raw `serde_json::from_value`.

For `{"vec": {"array": ["u8", 32]}}`, apply it per-element.

This sidesteps the need for the generator to distinguish "random nonce" from "account ID" (which the IDL type alone doesn't encode), and makes all 32-byte args uniformly user-friendly. Accepting base58 for a field like `create_key` is harmless — users just won't do it.

```rust
fn parse_bytes32(v: &Value) -> Result<[u8; 32], String> {
    if let Some(s) = v.as_str() {
        let s = s.strip_prefix("Public/")
            .or_else(|| s.strip_prefix("Private/"))
            .unwrap_or(s);
        // try base58 first
        use base58::FromBase58;
        if let Ok(bytes) = s.from_base58() {
            if bytes.len() == 32 {
                let mut arr = [0u8; 32];
                arr.copy_from_slice(&bytes);
                return Ok(arr);
            }
        }
        // fall through to hex
        let s = s.trim_start_matches("0x").trim_start_matches("0X");
        if s.len() == 64 {
            let bytes = hex::decode(s).map_err(|e| format!("invalid hex: {}", e))?;
            let mut arr = [0u8; 32];
            arr.copy_from_slice(&bytes);
            return Ok(arr);
        }
        return Err(format!("expected 32-byte base58, hex, or byte array, got: {:?}", v));
    }
    serde_json::from_value(v.clone()).map_err(|e| format!("parse error: {}", e))
}
```

Replace all `serde_json::from_value(v["field"]...)` for `[u8; 32]` instruction args with `parse_bytes32(&v["field"])?`.

vpavlin:
## Fix: `[u8; 32]` args now accept base58/hex (654f986)

The generated FFI already emitted a `parse_account_id` helper accepting base58 (`Public/`/`Private/` prefix), hex (`0x` prefix), and raw hex — the same logic as the SPEL CLI's `decode_bytes_32`. The bug was that `IdlType::Array { array: (Primitive("u8"), 32) }` (how `[u8; 32]` is represented when declared inline in the IDL) fell through to `serde_json::from_value` instead of being routed there.

Fix: one arm in `idl_type_to_json_parse` for `[u8; 32]` array types pointing to `parse_account_id`. `Vec<[u8; 32]>` gets it for free via the recursive Vec path.

**Before:** `parse error: invalid type: string "Public/CbgR6tj5kWx5...", expected a sequence`  
**After:** `members`, `new_member`, `member` all accept `Public/CbgR6...`, bare base58, `0x<hex>`, or raw hex — same as the CLI.

vpavlin:
## Additional issue: `member_accounts` requires a separate JSON field the UI doesn't send

When `create_multisig` is called, the generated FFI expects **two** separate JSON fields:
- `members` — the instruction arg (`Vec<[u8; 32]>`)
- `member_accounts` — the transaction account slots (`Vec<AccountId>`)

The QML UI only sends `members`. Since both fields contain the same keys (members are also the accounts that need to be listed in the NSSA transaction), the generated FFI should **derive `member_accounts` from `members`** instead of requiring it separately:

```rust
// Instead of:
let member_accounts: Vec<AccountId> = v["member_accounts"].as_array()
    .ok_or("missing member_accounts")?
    .iter().map(|a| parse_account_id(a.as_str().ok_or("expected string")?)).collect::<Result<Vec<_>,_>>()?;

// Generate:
let member_accounts: Vec<AccountId> = members.iter().map(|m| AccountId::new(*m)).collect();
```

More generally: when an instruction has a `Vec<[u8; 32]>` (or `AccountId`) arg whose values are also listed as non-signer/non-PDA accounts in the IDL, the generator should recognise the overlap and derive the account list from the parsed arg rather than requiring a duplicate field in the JSON payload.

vpavlin:
## Additional issue: no block-inclusion confirmation

The generated FFI calls `sequencer_client.send_transaction(...)` and immediately returns the tx hash as `success: true`. This means the UI shows success after the transaction is *submitted to the mempool*, not after it's *included in a block*.

The CLI uses `wallet::poller::TxPoller` to poll until the transaction is confirmed:

```rust
let raw_hash = wallet.sequencer_client.send_transaction(...).await
    .map_err(|e| format!("submit: {}", e))?;
let tx_hash_hex = hex::encode(raw_hash.0);
let poller = wallet::poller::TxPoller::new(wallet.config(), wallet.sequencer_client.clone());
poller.poll_tx(raw_hash).await.map_err(|e| format!("confirmation: {}", e))?;
Ok::<String, String>(tx_hash_hex)
```

The generated FFI should do the same — the current fire-and-forget pattern gives false confidence of success.

vpavlin:
**Correction to my previous comment** on `member_accounts`:

The CLI *does* require `member_accounts` separately — as a `--member-accounts` flag (it's a `rest` account in the IDL). The user passes the same keys twice:

```bash
spel -- create-multisig --members <k1>,<k2> --member-accounts <k1>,<k2>
```

So the generated FFI requiring `member_accounts` as a separate JSON field is technically consistent with the CLI. Our local workaround (deriving it from `members`) is a UX improvement for the UI case, not a correctness fix.

The real question for the generator: when a `rest` account list is identical in type and value to an instruction arg (both are `[u8;32]` keys representing the same accounts), should the generator emit a smarter pattern that derives one from the other? For a UI, asking users to type the same keys twice is untenable. For a CLI it's at least scriptable.

vpavlin:
## Update: `#[account_type]` state fetch works end-to-end

After adding `#[account_type]` to `MultisigState` and `Proposal` in `multisig_core` (a path-dep crate) and updating the IDL gen to use the commit with `find_path_dep_dirs`, both account types now appear in the IDL and `--target logos-module` correctly generates `fetchMultisigState` / `fetchProposal` Q_INVOKABLE methods and matching `Q_PROPERTY(QVariantMap ...)` properties.

### Two issues hit during integration

**1. `fetch_proposal` missing from the FFI (non-`logos-module` target)**

The default FFI target (`--target` not set) only emits a fetch function for `multisig_state`. It does _not_ emit one for `proposal`, even though `proposal` is in `idl.accounts`. This is because `proposal` is a PDA account whose PDA is computed from `create_key` + `proposal_index`, and the generator apparently skips it (or requires explicit seeds in the IDL to generate the fetch).

The `logos-module` Qt backend calls `multisig_program_fetch_proposal` from C++, but the symbol never exists in the compiled `.so`, causing a linker error.

Workaround: manually added `fetch_proposal` to the generated `multisig.rs` using the already-generated `compute_proposal_pda` helper, plus corresponding Borsh structs for `ProposalStatusState`, `ConfigActionState`, and `ProposalState`. Also added the C header declaration to `lez_multisig.h`.

The generator should emit a fetch function for every account in `idl.accounts`, regardless of whether it's a PDA or not. The function signature just needs additional seed arguments for PDA accounts — e.g. `fn multisig_program_fetch_proposal(args_json)` where args contains `create_key` and `proposal_index`.

**2. CMake resource name collision in standalone app**

The generated `LezMultisigPlugin.cpp` calls `Q_INIT_RESOURCE(lez_multisig_qml)`, but the standalone app target in `CMakeLists.txt` uses a _different_ resource name (`lez_multisig_app_qml`), causing an `undefined reference to qInitResources_lez_multisig_qml()` linker error.

Fix: rename the app target's `qt_add_resources(... "lez_multisig_app_qml" ...)` to `"lez_multisig_qml"` to match the plugin's resource name. The generator should emit a consistent resource name across both the plugin and standalone app targets (or document that the app must use the same name).

vpavlin:
## Additional issue: `ColumnLayout` padding properties crash QML (white screen)

The generated `Main.qml` sets padding directly on `ColumnLayout`:

```qml
ColumnLayout {
    width: scrollView.width
    spacing: 12
    leftPadding: 16; rightPadding: 16
    topPadding: 16; bottomPadding: 80
```

Two bugs here:

1. **Semicolon-separated property bindings on one line** (`leftPadding: 16; rightPadding: 16`) — the QML parser treats the second binding as a JavaScript label expression rather than a property assignment. Result: `Cannot assign to non-existent property "rightPadding"`.

2. **`bottomPadding` / `topPadding` not exposed on `ColumnLayout`** in some Qt builds — these are `Item` properties but `ColumnLayout`'s metatype doesn't always surface them. Result: same error, white screen instead of rendered UI.

Fix: use `x: 16; width: parent.width - 32` for horizontal margins, and `Item { height: N; Layout.fillWidth: true }` spacers for top/bottom padding — these work across all Qt 6 builds.

---

## Broader concern: generated flat-form UI doesn't fit stateful programs

The generated `Main.qml` + `LezMultisigBackend` is a flat scrollable list of all instruction forms with raw parameter fields. For `lez-multisig` this means the user must manually enter `proposalIndex`, `createKey`, `approverId`, etc. on every call — there's no state carried between operations.

The hand-crafted backend we had before exposed:
- `backend.multisigs` (locally-persisted list)
- `backend.currentMultisig` / `backend.proposals` (fetched on selection)
- `backend.selectMultisig(createKey)` / `backend.loadMultisig(createKey)`
- `backend.approve(index)` / `backend.execute(index)` — no need to re-enter createKey each time

The generated scaffold is a good starting point, but for programs with persistent accounts it's the wrong default UX. It would help if `--target logos-module` either:
- Generated a richer stateful backend when the IDL has `accounts` types, or
- Made it clear in a README/comment that the backend is intentionally minimal and meant to be extended

vpavlin:
## Follow-up 5: TxPoller, rest accounts, all-accounts fetch, CMakeLists (4b8a71c)

Four fixes in one commit:

### ✅ Block-inclusion confirmation (TxPoller)

Generated instruction FFI now polls for confirmation after submit:
```rust
let raw = wallet.sequencer_client.send_transaction(...).await
    .map_err(|e| format!("submit: {}", e))?;
let tx_hash_hex = hex::encode(raw.0);
let poller = wallet::poller::TxPoller::new(wallet.config(), wallet.sequencer_client.clone());
poller.poll_tx(raw).await.map_err(|e| format!("confirm: {}", e))?;
Ok::<String, String>(tx_hash_hex)
```
`operationSuccess` in the QML backend now fires only after block inclusion. The UI busy indicator stays active until then.

### ✅ `rest` accounts derived from Vec<[u8;32]> arg

When a `rest` account exists and the instruction has a `Vec<[u8;32]>` arg, the JSON field is now **optional** — if absent, rest accounts are derived from the parsed arg (same values, already typed as `Vec<AccountId>`). The UI sends keys once; the CLI can still pass `member_accounts` explicitly for backward compatibility.

### ✅ Fetch functions for all `idl.accounts` entries

`generate_account_fetch_functions` now iterates `idl.accounts` directly instead of instruction PDA slots. Every struct-typed account gets a fetch function:
- PDA accounts: seed computation as before
- Non-PDA accounts (e.g. `proposal`): accept `account_id` directly in the JSON args

Fixes the linker error where `lez_multisig_fetch_proposal` was declared in the C++ backend but never emitted by the FFI crate.

### ✅ Generated `CMakeLists.txt`

`--target logos-module` now emits a 10th file: `CMakeLists.txt`. Both the plugin and standalone app targets use the same `qt_add_resources` name (`lez_multisig`), so `Q_INIT_RESOURCE(lez_multisig)` in the plugin always resolves regardless of which target is being built.

vpavlin:
## New compile errors in generated FFI (after latest commit)

After regenerating with the `4b8a71c` tip, `cargo build -p lez-multisig-ffi` fails with 13 errors.

**1. `create_key` / `member` / `new_member` parsed as `AccountId` but instruction fields expect `[u8; 32]`**

The new generator emits:
```rust
let create_key = parse_account_id(v["create_key"].as_str().ok_or("expected string for [u8; 32]")?)?;
// ... later:
let instruction = ProgramInstruction::ProposeRemoveMember { member, create_key, proposal_index };
//                                                                   ^^^^^^^^^^ E0308: expected [u8; 32], found AccountId
```

`create_key`, `new_member`, and `member` are PDA-seed bytes (`[u8; 32]`), not account addresses. They need `parse_bytes32(&v["create_key"])?` (already present in the file as a helper), not `parse_account_id`.

The legitimate `AccountId` fields — `multisig_state`, `proposer`, `approver`, `rejector`, `executor` — are fine with `parse_account_id`.

**2. `ProposalState` struct not defined**

`fetch_proposal_impl` references `<ProposalState as borsh::BorshDeserialize>` but the struct is never emitted:
```
error[E0412]: cannot find type `ProposalState` in this scope
```

The generator emits the fetch function but not the matching Borsh deserialization struct. `MultisigStateState` (for the other account) is defined correctly, so this is a gap in the proposal account codegen path.

vpavlin:
## Bug: `member_accounts` rest-account else-branch emits a type error

In `ffi_codegen.rs`, when a `rest` account has a matching `Vec<[u8;32]>` instruction arg, the generator emits:

```rust
let member_accounts: Vec<AccountId> = if v["member_accounts"].is_array() {
    v["member_accounts"].as_array().unwrap()
        .iter().map(|a| parse_account_id(a.as_str().ok_or("expected string")?)).collect::<Result<Vec<_>,_>>()?
} else {
    members.clone()   // ← type error: members is Vec<[u8; 32]>, not Vec<AccountId>
};
```

The else-branch clones `members` (typed `Vec<[u8; 32]>`) into a binding typed `Vec<AccountId>` — these are different types and it won't compile.

**Fix:** change the else-branch to:
```rust
    {src_name}.iter().map(|m| AccountId::new(*m)).collect()
```

i.e. in `ffi_codegen.rs` around the `writeln!(out, "        {src_name}.clone()")` line, replace `.clone()` with `.iter().map(|m| AccountId::new(*m)).collect()`.

Currently patched in `lez-multisig` via `scripts/patch-ffi-gen.sh` as a workaround.

vpavlin:
Recommended changes before merge (non-blocking):                                                                                                                                                                                                                                                                                                                     
 1. Tweak the PDA dedup test assertion to be more precise                                                                                                                                                                                                                                                                                                             
 2. Consider whether the decode module should be feature-gated in spel-framework-core                                                                                                                                                                                                                                                                                 
 3. Normalize indentation in logos_module_codegen.rs

vpavlin:
## Follow-up: PR review comments addressed (65a0e29 + d1057e7)

**Fixed in `65a0e29`:**

1. **`{effective_prog}_program_id()` symbol mismatch** — `program_id()` now uses `effective_prog` (the `--module-name` value) instead of `prog` (IDL name), so it matches the FFI symbol users are expected to export. `free_string` keeps `prog` since it's a standard generated symbol.

2. **`i64` overflow on values > i64::MAX** — The u64 fallback path now rejects values > `i64::MAX` with an error instead of silently wrapping to negative.

3. **README env-var name** — Updated `{PROGRAM}_PROGRAM_ID_HEX` → `{PROGRAM}_PROGRAM_ID` in both the quick-start example and the Configuration section to match the generated backend code.

4. **`make ui-run` hardcoded executable** — Replaced `$(UI_OUT_DIR)/build/{pascal_name}App` with `find $(UI_OUT_DIR)/build -maxdepth 1 -name '*App' -type f` so it works when the CMake class name differs from the project name. Also removed now-unused `pascal_name` variable.

5. **`test-modules/Makefile` developer path** — Changed `SPEL_CLIENT_GEN` default from an absolute `/home/vpavlin/...` path to `$(shell which spel-client-gen 2>/dev/null || echo ../target/release/spel-client-gen)`.

**Fixed in `d1057e7`:**

6. **Raw string literal for embedded IDL JSON** — Replaced `r####"..."####` with an escaped regular string literal so any IDL content (including one that contains `"####`) produces valid Rust output.

**Not addressed (non-blocking / design decisions):**

- `decode_account_data` trailing-byte detection — design-level change, left to a separate PR
- PDA dedup test precision — existing `count == 1` assert already checks for exactly one occurrence
- `decode` module feature-gating — design decision for spel-framework-core owners
- indentation normalization in `logos_module_codegen.rs` — happy to do in a follow-up cleanup PR if wanted
- `sed` in lez-compat workflow — the current `\"` escaping inside double-quoted bash strings is valid: bash converts `\"` to `"` before passing to sed, so `[^"]*` is correct sed syntax

PR #232, opened 2026-06-16
fix(idl): emit account helper types in deterministic order

0x-r4bbit:
`collect_account_types` seeded its BFS queue via a `HashSet`, whose iteration order is randomized per process. The order helper types were discovered — and emitted in the IDL — therefore varied run-to-run, so the same source produced byte-different IDL.

Use order-preserving dedup and sort the returned helper types by name, so output is byte-stable and independent of source declaration order. This is shared by both the CLI and proc-macro paths. Add a regression test.

### DIFF
--- a/spel-framework-core/src/account_types.rs
+++ b/spel-framework-core/src/account_types.rs
@@ -232,13 +232,19 @@ pub fn collect_account_types(items: &[Item]) -> (Vec<IdlAccountType>, Vec<IdlTyp
     let mut helper_types: Vec<IdlTypeDef> = Vec::new();
     let mut visited: HashSet<String> = annotated_names.clone();
 
-    let mut queue: Vec<String> = accounts
-        .iter()
-        .flat_map(|a| collect_defined_refs(&a.type_))
-        .filter(|n| !visited.contains(n))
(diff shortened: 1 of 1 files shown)


vpavlin (approved):
Reviewed — approving.

Fixes a real determinism bug: `collect_account_types` seeded its BFS queue via a `HashSet` round-trip, whose per-process randomized iteration order made the emitted IDL byte-different across runs (breaks committed-IDL CI checks and produces spurious diffs).

Traced the function: the *set* of helper types is already order-independent (BFS reaches all reachable `Defined` types; `visited` guarantees each is emitted once) — only the order varied. The actual guarantee is the new name-sort at the end; type names are unique so it's total and deterministic. Good regression test (helpers referenced out of alphabetical order, asserts sorted output).

Non-blocking notes:
- The order-preserving-dedup rewrite of the initial queue is redundant with the final sort (the sort alone suffices), but it's correct and clearer — no change requested.
- Helper types are now alphabetical rather than topological; fine for IDL consumers and strictly better than the previous random order.

Verified against the LEZ v0.2.0 base after update-branch.

PR #180, opened 2026-05-06
fix(idl-gen): scan path-dep crates in generate_idl! macro + support qualified attribute form (issue #176)

vpavlin:
## Summary

Fixes both root causes identified in [issue #176](https://github.com/logos-co/spel/issues/176).

**Design:** path-dep scanning functions were moved from spel-cli into \`spel-framework-core\` so **both** the CLI and the proc-macro share a single 1:1 implementation. A generic \`on_warning: F: FnMut(String)\` callback parameter lets each caller choose its own error handling — CLI collects warnings for user display, proc-macro passes \`|_| {}\` (no-op).

### 1. Path-dependency crate scanning in \`generate_idl!\` macro

The \`generate_idl!\` proc macro now scans local path-dependency crates for \`#[account_type]\`-annotated types, matching the behavior that was already present in the CLI (\`spel generate-idl\`) since PRs #169 and #175.

**How it works:**
- Path-dep scanning functions moved from spel-cli to \`spel-framework-core/src/idl_gen.rs\`
- Both CLI and proc-macro call \`spel_framework_core::idl_gen::find_path_dep_dirs()\` with the same code path
- At macro expansion time, the nearest \`Cargo.toml\` is found by walking up from the source file
- Parses \`[dependencies]\` for \`path = "..." entries (excluding dev/build deps, registry deps, and git deps)
- **Workspace support:** detects virtual workspace roots and searches member directories
- **Transitive dependencies:** follows path-based dependencies through multiple levels with cycle detection

### 2. Qualified attribute form recognition

\`has_account_type_attr()\` now matches both:
- Bare form: \`#[account_type]\`
- Fully-qualified form: \`#[spel_framework_macros::account_type]\` (idiomatic when importing via a path rather than a \`use\` declaration)

This fix applies to both the CLI and proc-macro paths since they share the same \`has_account_type_attr()\` function.

### Changes

| File | Change |
|------|--------|
| \`spel-framework-core/Cargo.toml\` | Added \`toml\` as optional dependency behind \`idl-gen\` feature |
| \`spel-framework-core/src/idl_gen.rs\` | Added shared \`find_path_dep_dirs()\` + helpers with generic warning callback; made \`collect_items_from_crate_dirs()\` public |
| \`spel-framework-core/src/account_types.rs\` | Updated \`has_account_type_attr()\` to match qualified paths; changed visibility from \`pub(crate)\` to \`pub\` |
[...]

PR #233, opened 2026-06-24
RFP-001 framework enabler: generic `[package.metadata.spel]` extension scanner

mmlado:
## Description

> **Note on RFP-001 deliverables.** This PR is the framework-side enabling change for [RFP-001, awarded to @mmlado](https://github.com/logos-co/rfp/issues/46). The RFP-001 deliverables themselves (admin-authority library, sample program, integration tests, ADRs, lifecycle docs) live in the companion repository [`mmlado/spel-admin-authority`](https://github.com/mmlado/spel-admin-authority), per the contract's primary-repo routing. This framework PR carries no admin-specific code; it adds a generic extension scanner that admin-authority is the first consumer of. The two should be reviewed together but ship under separate review tracks. PR for `admin-authority` M1 can be found in its repository: [RFP-001 Milestone 1, macro API design, account model, framework integration path](https://github.com/mmlado/spel-admin-authority/pull/1)

Adds a generic extension-discovery mechanism to `spel-framework-core` + `spel-framework-macros`. Path-dep libraries that declare `[package.metadata.spel]` in their `Cargo.toml` can ship `#[instruction]` fns that are auto-discovered by the framework and merged into a consuming program's dispatcher + IDL.

The framework has **zero hardcoded knowledge** of any specific extension library. RFP-001's `admin-authority` is the first consumer of this mechanism. RFP-002's `freeze-authority` (and any future SPEL extension) plug in by declaring the same metadata, no further framework changes required.

Motivation: without this mechanism, every new SPEL extension would require a framework PR carrying library-specific code (synthetic instruction templates, hardcoded attr names, library-side imports). That coupling does not scale across multiple libraries and locks each extension to the framework's release cadence. Metadata-driven discovery decouples the two: libraries ship independently, framework stays library-agnostic.

## Changes

- **`spel-framework-core::idl_gen`** (behind `idl-gen` feature): five new public helpers for metadata reading + extension discovery (`read_spel_extension_attr`, `read_spel_instruction_attrs`, `discover_extension_instructions`, `discover_extension_instruction_attrs`, `collect_instruction_fns`).
- **`spel-framework-macros::expand_lez_program`**: calls discovery after parsing user instructions; appends discovered fns tagged with `external_call_path`; passes collected `instruction_attrs` to handler emit for stripping.
- **`spel-framework-macros::expand_generate_idl`** and **`spel-framework-core::idl_gen::generate_idl_inner`**: same discovery loop wired in so `make idl` and `spel-cli generate-idl` match compile-time codegen.
- **`InstructionInfo`** struct: new `external_call_path: Option<syn::Path>` field. `Some(path)` triggers cross-crate dispatcher call + handler-emit skip.
- **`generate_handler_fns`** + **`generate_match_arms`**: respect `external_call_path` (skip handler emit, emit cross-crate dispatch); strip per-extension instruction attrs.
- **`README.md`**: new "Extension Libraries" section describing the discovery mechanism for library authors.
- **6 new unit tests** in `spel-framework-core::idl_gen::tests` covering metadata reads + discovery filtering.

## How it works

1. Extension library declares discovery metadata:

   ```toml
   # admin-authority/Cargo.toml
   [package.metadata.spel]
   extension_attr = "admin_authority"
   instruction_attrs = ["require_admin"]
   ```

2. Consumer puts the marker on a `#[lez_program]` module:

   ```rust
   #[lez_program]
   #[admin_authority]
   mod my_program { /* user's own instructions */ }
   ```

3. Framework scans path-deps, finds matching metadata, parses the library's `src/lib.rs` for `#[instruction]` fns, and merges them into the consumer's pipeline with cross-crate dispatcher calls (`::admin_authority::admin_initialize(...)`).

Existing path-dep scanning machinery from [#180](https://github.com/logos-co/spel/pull/180) is reused; this PR adds three sibling helpers and one signature change.

### `spel-framework-core::idl_gen` (helpers, behind `idl-gen` feature)

- `read_spel_extension_attr(crate_dir) -> Option<String>`, read `[package.metadata.spel.extension_attr]`.
- `read_spel_instruction_attrs(crate_dir) -> Vec<String>`, read `[package.metadata.spel.instruction_attrs]`.
- `discover_extension_instructions(manifest_dir, mod_attrs) -> Vec<(syn::ItemFn, syn::Path)>`, orchestrator: walk path-deps, filter by attr match, return (fn, crate_path) pairs.
- `discover_extension_instruction_attrs(manifest_dir, mod_attrs) -> Vec<String>`, orchestrator: collect declared instruction-level marker names so the framework can strip them from emitted handler fns.
- `collect_instruction_fns(items) -> Vec<syn::ItemFn>`, filter helper used by the discover orchestrator.

### `spel-framework-macros::expand_lez_program`

- Calls `discover_extension_instructions` after parsing user-defined instructions; appends each with `external_call_path = Some(::<crate>::<fn>)`.
- Calls `discover_extension_instruction_attrs`; passes the list into `generate_handler_fns` to strip those attrs before emit (prevents re-expansion of library-owned gate macros on the emitted handler).

### `InstructionInfo`

- New field: `external_call_path: Option<syn::Path>`.
  - `None` → dispatcher emits bare-name call (current behavior).
  - `Some(path)` → dispatcher emits cross-crate call; `generate_handler_fns` skips emit (the body lives in the library).

### `spel-framework-macros::expand_generate_idl` and `spel-framework-core::idl_gen::generate_idl_inner`

- Same discovery loop wired in, so `make idl` and `spel-cli generate-idl` pick up extension instructions identically to compile-time codegen.

### What is **not** added

No admin-specific code in the framework. No `admin_authority` module. No `require_admin` field. No library-specific attribute strings.

## Companion repository

The runtime library that this scanner discovers lives in a separate repository: [`mmlado/spel-admin-authority`](https://github.com/mmlado/spel-admin-authority). It contains:

- `admin-authority/`, runtime crate: `AdminConfig`, `AdminCandidate`, `AdminError`, the three `#[instruction]` fns, and `[package.metadata.spel]` declarations.
- `admin-authority-macros/`, proc-macro sub-crate: `#[admin_authority]` (marker), `#[require_admin]` (shape validator), `#[instruction]` shim.
- `admin-authority-sample/`, reference SPEL program consuming both macros end-to-end.

This PR can be reviewed independently of the companion repo. Library method bodies (currently `todo!()` stubs) and the runtime auth check inside `#[require_admin]` land in M2.

## Test coverage

Added unit tests in `spel-framework-core::idl_gen::tests`:

- `read_spel_extension_attr_returns_declared_name`
- `read_spel_extension_attr_none_when_missing`
- `read_spel_extension_attr_none_when_no_manifest`
- `discover_extension_instructions_picks_up_matching_ext`
- `discover_extension_instructions_skips_when_attr_absent_on_mod`
- `discover_extension_instructions_skips_deps_without_metadata`

Existing path-dep test infrastructure (`find_path_dep_dirs_*`) is reused. End-to-end verification via the companion repo's `admin-authority-sample/tests/idl.rs` integration test, which runs the full pipeline (sample source → discovery → IDL JSON containing user instr + admin trio).

Run with:

```bash
cargo test -p spel-framework-core --features idl-gen
cargo test -p spel-framework-macros
```

## Design notes

- **Per-crate dep declaration is unavoidable.** Cargo metadata is a discovery hint, not a dep injection mechanism. Consumers still add `admin-authority = "..."` (or any future extension) to their `Cargo.toml`; metadata only tells the framework what to scan once the dep is present.
- **Re-expansion avoidance.** Library-owned gate macros (e.g. `#[require_admin]`) attached to a user fn would otherwise re-expand on the handler emitted by `#[lez_program]`. Stripping is driven by `instruction_attrs` metadata so the framework stays generic; library authors declare which attrs they own.
- **Stacking.** Multiple extensions on one consumer module (`#[admin_authority] #[freeze_authority]`) work without cross-extension coordination, each discovers independently, each declares its own `instruction_attrs`.
- **Self-reference in library source.** Library instruction fns reference their own crate's types via the absolute path (`::admin_authority::AdminCandidate`). The library's `src/lib.rs` adds `extern crate self as admin_authority;` to make that path resolve in the library's own compile.

## Open questions for maintainers

1. **Metadata namespace.** `[package.metadata.spel]` chosen for brevity. Acceptable, or prefer `[package.metadata.spel-extension]` / `[package.metadata.spel.discovery]` / similar?
2. **Field names.** `extension_attr` (singular) + `instruction_attrs` (plural). Open to alternatives, `module_attr` / `gate_attrs`, etc.
3. **Scope of discovery.** Currently scans only local path-dep crates (matches `#[account_type]` scanning behavior from #180). Should registry / git deps be scanned too, given the trust model is already opt-in via metadata? Default leaning is "no, keep symmetry with existing behavior".
4. **Crate dir → Rust ident.** The discovered cross-crate path uses `dep_dir.file_name().replace('-', '_')`. Robust against the common `kebab-case` crate convention; could be brittle if a library declares `package.name` different from its directory. Switch to reading `[package].name` from the manifest if maintainers prefer.

## Breaking changes

None. Purely additive. Consumers without `[package.metadata.spel]` in any path-dep, or without an extension attr on their `#[lez_program]` module, see zero behavior change.

## M1 deliverables and completion criteria

Per the [proposal](https://github.com/logos-co/rfp/issues/46):

**Deliverables**

- [x] `#[require_admin]` macro API design, see [companion repo `docs/authority-lifecycle.md`](https://github.com/mmlado/spel-admin-authority/blob/main/docs/authority-lifecycle.md) and `docs/adr/0003-require-admin-strict-by-default.md`.
- [x] `AdminConfig` account model, defined in [`admin-authority/src/lib.rs`](https://github.com/mmlado/spel-admin-authority/blob/main/admin-authority/src/lib.rs); seed choice in `docs/adr/0001-admin-config-pda-seed.md`.
- [x] Authority lifecycle documentation, `docs/authority-lifecycle.md` covers states, transitions, validation rules, signer/PDA paths, and the program-as-admin CPI flow.
- [x] Draft PR opened to `logos-co/spel`, this PR.
- [x] LEE toolchain setup, `docs/toolchain-and-macros-study.md` records tooling versions, build modes, verification steps.
- [x] `spel-framework-macros` codebase study, same document maps entry points, internal seams, and patterns this integration follows.

**Completion criteria**

- [x] Draft PR open on fork branch (`mmlado/spel:feat/admin_authority`), ready for upstream review, this PR.
- [x] Generic extension discovery mechanism documented and ready for maintainer review, see Summary, How it works, Open questions.
- [x] `AdminConfig` design documented, `admin-authority/src/lib.rs` and the lifecycle doc; on-chain encoding is a single 32-byte `AccountId` with `AccountId::default()` as the renounced sentinel.

## Checklist

- [x] Builds cleanly (`cargo build --workspace --features idl-gen`)
- [x] Tests pass (`cargo test --workspace --features idl-gen`)
- [x] README updated (new "Extension Libraries" section in `README.md` describing the discovery mechanism)
- [x] New public methods have doc comments (`read_spel_extension_attr`, `read_spel_instruction_attrs`, `discover_extension_instructions`, `discover_extension_instruction_attrs`, `collect_instruction_fns`)
- [x] Branch is off `main` (forked from `73fc462`, current `logos-co/spel:main` HEAD at branch creation)


mmlado:
While working on M2 of admin-authority (RFP-001) I needed a way to make sure a new admin key is real and controlled by its holder. The only proof of key control is a signature made with that key. There are two ways to get one:

- a two-step transfer, where the new admin accepts a nomination in its own on-chain transaction
- a multi-signature transaction, where the new admin co-signs the transfer itself

spel-cli already supports multi-signature transactions, but only when all signers live in the same wallet. I built on that. The current admin prepares a transaction, saves it to a file, and hands it to the new admin. The new admin signs the same transaction, and once all signatures are gathered any party can submit it to the sequencer.

The file format is deliberately tool agnostic (JSON over borsh nssa types), so the concept could live in nssa long term. It works with any number of co-signers and should be useful well beyond this use case.

Since it stands on its own, it is a separate PR: #246. The admin-authority transfer to a keyholder (the Signer candidate) relies on it and is not usable on chain without it.

spel-framework-core/src/extension.rs
```diff
@@ -0,0 +1,316 @@
...
+        };
+        if !mod_attrs.iter().any(|a| a.path().is_ident(&ext_attr)) {
+            continue;
+        }
+
+        let (items, _) = collect_items_from_crate_dirs(&[dep_dir.clone()]);
+        let Some(crate_name) = dep_dir.file_name().and_then(|n| n.to_str()) else {
+            continue;
+        };
+        let crate_name = crate_name.replace('-', "_");
```
danisharora099: The cross-crate call path is derived from the dependency's directory name here, not from `[package].name`. It works in the current setup only because admin-authority's directory happens to match its package name; a renamed, vendored, or `package =` aliased dep emits a call to a nonexistent module path and the consumer fails to compile.

One important request on sequencing: please land the fix **together with** a consumer-side opt-in/allow-list. Today a registry/git crate declaring a matching `extension_attr` breaks the build (this bug); with the naive fix alone it would instead get its instructions silently injected into the consumer's dispatcher.

spel-framework-core/src/extension.rs
```diff
@@ -0,0 +1,316 @@
...
+}
+
+/// Scan path-dep crates for SPEL extension libraries whose `extension_attr`
+/// metadata matches an attribute on the consuming program's mod. Returns
+/// one `(ItemFn, crate_path)` per discovered `#[instruction]` fn, where
+/// `crate_path` is the absolute path to call the fn from the consumer
+/// (e.g. `::admin_authority`).
+///
+/// Silent on env/IO errors — returns empty. Callers may emit their own
+/// error if they expected results.
```
danisharora099: This makes discovery fail open with no diagnostics: a mis-set `CARGO_MANIFEST_DIR`, an unreadable/malformed manifest, or a typo'd extension attr yields a program that builds and deploys without its admin surface. For a security-library mechanism that's the wrong default. #180 added an `on_warning` channel for exactly this; please wire it through here (the CLI already surfaces warnings).

spel-framework-core/src/extension.rs
```diff
@@ -0,0 +1,316 @@
...
+pub fn discover_extension_instructions(
+    manifest_dir: &Path,
+    mod_attrs: &[syn::Attribute],
+) -> Vec<(syn::ItemFn, syn::Path)> {
+    let dep_dirs = find_path_dep_dirs(&manifest_dir, |_| {});
```
danisharora099: Trust-model question that needs an ADR before M2 logic rides on this: discovery walks transitive path-deps and keys on an attr-name match only, so a transitive dep the consumer never opted into can contribute instructions to the dispatcher if it declares the same `extension_attr`. Suggest binding discovery to crate identity (name/version, or an explicit allow-list on the consumer module) and/or restricting to direct deps.

spel-framework-macros/src/lib.rs
```diff
@@ -1162,12 +1181,19 @@ fn extract_vec_macro_idents(expr: &syn::Expr) -> Option<Vec<Ident>> {
...
+) -> Vec<TokenStream2> {
     instructions
         .iter()
+        .filter(|ix| ix.external_call_path.is_none())
         .map(|ix| {
             let mut func = ix.func.clone();
             func.attrs.retain(|a| !a.path().is_ident("instruction"));
+            for name in strip_attrs {
+                func.attrs.retain(|a| !a.path().is_ident(name))
+            }
```
danisharora099, line 1223: I think the "prevent re-expansion" rationale (extension.rs:68-69, README:228) mischaracterizes what this loop does. Attribute macros on items inside a module only expand after the outer `#[lez_program]` rewrite, so stripping here prevents the first (and only possible) expansion, not a re-expansion. The upshot is that a gate attr on a consumer instruction contributes no code, no validation, and no diagnostics. That may well be the intended design, but it should be documented explicitly (library gates apply via re-expansion on library-emitted handlers only) so extension authors don't rely on consumer-side gating that never runs.

spel-framework-macros/src/lib.rs
```diff
@@ -241,6 +242,16 @@ fn expand_lez_program(input: ItemMod, config: ProgramConfig) -> syn::Result<Toke
...
     }
 
+    let manifest_dir = std::env::var("CARGO_MANIFEST_DIR")
+        .map_err(|_| syn::Error::new_spanned(&input.ident, "CARGO_MANIFEST_DIR not set"))?;
+    let manifest_dir = std::path::PathBuf::from(manifest_dir);
+    for (func, crate_path) in spel_framework_core::extension::discover_extension_instructions(&manifest_dir, &input.attrs) {
+        let mut info = parse_instruction(func)?;
+        let name = &info.fn_name;
+        info.external_call_path = Some(syn::parse_quote!(#crate_path::#name));
+        instructions.push(info);
```
danisharora099, line 258: No duplicate-instruction-name detection along this path: two extensions (or an extension and a user fn) sharing a name produce duplicate enum variants/match arms or silent shadowing, plus SHA256 discriminator collisions in the IDL. A clear compile-time error here (name -> first-seen source) would close it. Same applies to the parallel pushes in expand_generate_idl (line 2110) and generate_idl_inner.

README.md
```diff
@@ -215,6 +215,20 @@ Each account field includes:
...
+
+Third-party libraries can ship `#[instruction]` fns that are auto-discovered by the framework and merged into a consuming program's dispatcher and IDL. Discovery is driven by metadata in the library's `Cargo.toml`:
+
+```toml
+[package.metadata.spel]
+extension_attr = "admin_authority"
+instruction_attrs = ["require_admin"]
+```
+
+When a consumer's `#[lez_program]` module carries the declared `extension_attr` (e.g. `#[admin_authority]`), the framework scans the library's `src/lib.rs` for `#[instruction]` fns and merges them with cross-crate dispatcher calls. The `instruction_attrs` list names per-instruction marker attributes the library owns (e.g. `#[require_admin]`) so the framework can strip them from emitted handler fns to prevent re-expansion.
```
danisharora099: Three consumer-facing contracts this section should document: (1) extension instruction fns must be re-exported at the extension crate root or the generated cross-crate dispatcher call won't resolve; (2) extension signature types are copied verbatim into consumer-side codegen and must resolve at the consumer's expansion site; (3) gate attrs on consumer instructions are stripped and never expand (see the comment on the strip loop in spel-framework-macros).

danisharora099 (commented):
Thanks for this. I've done a full review of the extension-scanner mechanism (verified at head 95c1b81; framework tests green locally: 94+22 core w/ idl-gen, 10 macros, 51 client-gen). The generic design is sound and I support the direction; deliberately keeping admin-specific code out of the framework is the right call.

RFP-001 M1 verdict is on the companion library PR (mmlado/spel-admin-authority#1): accept with conditions. For this PR, the items below (inline) need addressing before it lands upstream. None of them change my view that this is the right architecture for extensions.

Housekeeping: `cargo fmt --check` currently fails (12 diffs), so the inherited quality.yml will red-flag this PR the moment CI runs, plus there are 3 lib-level unused-import warnings.

spel-framework-core/src/extension.rs
mmlado: Fixed in [a85c90c4799d8b5a1438c5dd802d81a093bacf4a](https://github.com/logos-co/spel/pull/233/changes/a85c90c4799d8b5a1438c5dd802d81a093bacf4a) . Identity now comes from [package].name. Discovery is restricted to direct dependencies. Opt-ins are explicit by the consumer this way: the dependency in the consumer's `Cargo.toml` and the marker attr on the module. A transitive crate claiming same extension_attr contributes nothing, test covers case.

spel-framework-core/src/extension.rs
mmlado: Fixed in [8b71d8a](https://github.com/logos-co/spel/pull/233/changes/8b71d8af31e8b2930332d423fdd404b10dbaad8b). Two tiers: malformed `[package.metadata.spel]` (wrong-shaped extension_attr or instruction_attrs) hard compile error and on_warning channel environmental skips (unreadable manifest, missing dep dir, a matched extension contributing neither instructions nor gate attributes). Docs updated to reflect this.

spel-framework-macros/src/lib.rs
mmlado: Fixed in [de465ac](https://github.com/logos-co/spel/pull/233/changes/de465ace5886ab865530c3fa37d3761f1e6d24e9). All three assembly sites run a shared first seen name check after merging the instructions. A collision fails compilation with both sources named, user vs extension and extensions between themselves, user vs user was already done by rustc.
Writing the test for this also exposed that [a85c90c](https://github.com/logos-co/spel/pull/233/changes/a85c90c4799d8b5a1438c5dd802d81a093bacf4a) had a broken discovery through the source scanner producer (file path vs crate dir), fixed in the same commit.

spel-framework-core/src/extension.rs
mmlado: Documented in [b4e0e98](https://github.com/logos-co/spel/pull/233/changes/b4e0e987de93592e3e83e054cbf34ebd2f1c68ac): docs/extension-trust-model.md. Records the opt-in mechanism: direct dependency plus module marker, package name identity. The code landed in [a85c90c](https://github.com/logos-co/spel/pull/233/changes/a85c90c4799d8b5a1438c5dd802d81a093bacf4a).

spel-framework-macros/src/lib.rs
mmlado: You're right. The strip prevents the first and only possible expansion, not a re-expansion. Fixed the rationale in the module docs, at the strip loop, and in the README. It now states that library gates take effect only by re-expanding on library-emitted handlers, and that a gate attr on a consumer-authored instruction contributes nothing (b4e0e98)

README.md
mmlado: Documented in b4e0e98. Crate-root re-export, absolute-path signature types, no consumer-side gate expansion, plus the trust model, in the Extension Libraries section.

danisharora099 (commented):
Took another pass at [`mmlado/spel@b4e0e98`](https://github.com/mmlado/spel/commit/b4e0e987de93592e3e83e054cbf34ebd2f1c68ac). All six items are addressed. The one I checked most carefully was the discovery change: the package-name fix and the direct-deps restriction landed in the same commit, which was the condition, since either alone would have left the silent-injection path open. The trust-model ADR matches the code, malformed metadata is a hard compile error, duplicate instruction names error naming both sources, and the extension-author contracts are in the README. fmt passes on my machine.

Nothing left from my side for the RFP milestones. For upstream landing, the open item is CI actually running on this PR, which is an M3 item.

spel-framework-core/src/idl_gen.rs
```diff
@@ -134,6 +150,33 @@ fn generate_idl_inner(
         return Err(IdlGenError::NoInstructions(path_str));
     }
 
+    if let Some(manifest_dir) = manifest_dir {
```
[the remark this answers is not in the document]
mmlado: This is by design. I don't see a point in a program that doesn't have any functions. The extensions functions are addons/utilities not program functions.
 @danisharora099 Do you think it has merit to have a program with only extension functions?

mmlado:
Superseded by https://github.com/logos-co/spel/pull/257, the single merge to main per the release agreement. The review history stays here.

Issue #239, opened 2026-06-29
Feature Request: Dynamic authorization requirements to SPEL


3esmit:
## Problem

SPEL currently models account authorization with one field:

```rust
#[account(signer)]
```

This maps to `accounts[].signer = true` and means the account must always be
authorized. That works for fixed signers. It fails for instructions where the
required signer depends on arguments, account state, delegation, smart-account
policy, or capability ownership.

Today there are two bad options:

- mark every possible signer as required, which over-constrains valid
  transactions;
- mark none as required, which leaves IDL-driven clients without signer, nonce,
  or witness guidance.

`signer: true` should keep meaning "always required." Dynamic authorization
needs separate metadata.

Currently we require to implement a more complex logic to circumvent this, see https://github.com/logos-blockchain/lez-programs/pull/208

## Proposal

Add instruction-level authorization requirements to IDL.

An authorization requirement describes:

- `id`: stable requirement id;
- `resources`: accounts/resources being authorized;
- `condition`: when the requirement applies;
- `satisfy`: witness expression that can satisfy it;
- `payload`: typed signing/replay metadata.

This separates account layout from authorization policy:

- account metadata says which accounts are passed and how they are accessed;
- authorization metadata says which authority role may need to be satisfied.

Runtime/program validation remains source of truth. IDL gives CLI and generated
clients enough information to collect the right witnesses.

## Macro sketch

Exact syntax is open. Shape should be instruction-scoped:

```rust
#[authorization(
    id = "selected_debit",
    resource(account = "source_a", access = "debit"),
    resource(account = "source_b", access = "debit"),
    when(runtime),
    satisfy(any(
        signed(account = "source_a", nonce = "source_a"),
        signed(account = "source_b", nonce = "source_b")
    )),
    payload(type = "SelectedDebitAuthorization")
)]
pub fn swap_or_route(
    #[account(mut)]
    source_a: AccountWithMetadata,
    #[account(mut)]
    source_b: AccountWithMetadata,
    #[account(mut)]
    recipient: AccountWithMetadata,
    selected_asset: AccountId,
) -> SpelResult
```

Generated validation must not require both sources to sign. Preflight or
program logic reports the selected source when authorization is missing.

The macro should validate referenced accounts and arguments exist.

## IDL sketch

Do not change existing account fields:

```json
{
  "name": "source_a",
  "writable": true,
  "signer": false,
  "init": false
}
```

Add optional authorization metadata to instructions:

```json
{
[...]

Issue #234, opened 2026-06-26
Public transactions built by SPEL are rejected on LEZ testnet with `InvalidSignature`


3esmit:
# Summary

SPEL can build and deploy a program to `https://testnet.lez.logos.co`, but public instruction transactions built by the current SPEL transaction path are rejected by the live testnet sequencer with:

```text
InvalidSignature
```

Deployment is not enough to catch this because `ProgramDeploymentTransaction` does not carry public-account signatures. The failure appears when submitting a public transaction with signer accounts.

The likely cause is a signature-preimage mismatch between the LEZ dependency currently pinned by SPEL and the signature scheme accepted by the live testnet.

Current SPEL `main` / `v0.5.0` pins `logos-execution-zone` `v0.1.2`, which resolves to commit `cf3639d8`. The SPEL CLI and generated clients construct public transactions through `WitnessSet::for_message(...)`:

```rust
let message = Message::new_preserialized(program_id, account_ids, nonces, instruction_data);
let witness_set = WitnessSet::for_message(&message, &signing_keys);
let tx = PublicTransaction::new(message, witness_set);
```

This path is present in:

- `spel-cli/src/tx.rs`
- `spel-client-gen/src/codegen.rs`
- `spel-client-gen/src/ffi_codegen.rs`

Live testnet accepts the same public transaction envelope when the public message is signed over:

```text
SHA256("/LEE/v0.3/Message/Public/\0\0\0\0\0\0\0" || borsh(public_message))
```

That matches current `logos-execution-zone` `main` behavior in:

- `lee/state_machine/src/public_transaction/message.rs`
- `lee/state_machine/src/public_transaction/witness_set.rs`
- `lee/state_machine/src/signature/mod.rs`

# Reproduction

1. Configure wallet for live LEZ testnet:

   ```bash
   wallet config set sequencer_addr https://testnet.lez.logos.co
   ```

2. Deploy a guest binary:

   ```bash
   wallet deploy-program target/riscv32im-risc0-zkvm-elf/docker/<program>.bin
   ```

   Deployment exits successfully and can be verified with sequencer RPC:

   ```bash
   curl -sS -X POST https://testnet.lez.logos.co/ \
     -H 'content-type: application/json' \
     --data '{"jsonrpc":"2.0","id":1,"method":"getTransaction","params":["<deploy-tx-hash>"]}'
   ```

3. Submit a public instruction with one or more signer public accounts through SPEL:

   ```bash
   spel \
[...]

PR #246, opened 2026-07-06
feat(cli): witness exchange for multi-signature transactions

mmlado:
## Description

Some instructions need signatures from accounts whose keys live on different machines. A transaction's authorization per account is derived from the witness set over the message, not from the `#[account(signer)]` annotation alone, and [admin-authority](https://github.com/mmlado/spel-admin-authority)'s `admin_transfer` with a `Signer` candidate is the first concrete case: the current admin and the new admin must both sign one message. The CLI could only sign with locally held keys.

This PR adds an offline partial-transaction flow. The exporting machine builds and signs the transaction and writes it to a file instead of submitting. The file travels to each co-signer over any channel, they inspect and sign it, and anyone submits once all witnesses are collected.

The file format is a thin JSON envelope over borsh nssa types, documented in `spel-cli/src/blob.rs`. Any tool can produce or consume it. The signers list is the order contract: it matches the nonce order inside the message and dictates witness order at submit. Verification checks each signature over the stored message bytes and checks that each pubkey derives the claimed account id.

Verified live on a local LEZ v0.1.2 stack with two wallets: export by one, sign by the other, submit passed signature and nonce validation and failed only at the expected Unknown program for an undeployed program id.

Related issues:

- #239 is the declaration half of the same problem: IDL metadata for conditional signers. This PR is the transport half. With #239 metadata the `--co-signer` flag could later be derived from the IDL and the instruction arguments.
- #234: while testing this feature against a local v0.1.2 stack, spel-built public transactions passed signature validation. The InvalidSignature rejection did not reproduce locally, see comment on the issue.

Known limits, deliberate for v1: co-signers decide based on the embedded summary (trust model documented in blob.rs, independent IDL decode of the message bytes is the planned upgrade). `--dry-run` does not show co-signers. `--export` supports public transactions only.

## Changes

- `spel-cli/src/blob.rs` (new): partial-transaction file format v1 with load/save, verification, and witness set assembly in signers order
- `--co-signer <id>` (repeatable) and `--export <file>` flags: build the message once with nonces for all signers, sign with wallet-held keys, write the file instead of submitting
- `spel sign <file>` (new subcommand): verify collected witnesses, show the embedded summary next to fields decoded locally from the message bytes, confirm, sign for wallet-held keys, write back. Works offline
- `spel submit <file>` (new subcommand): completeness check, witness set assembly in signers order, submit through the existing sequencer client
- Dry-run text renderer refactored to return a string so the export path embeds the same summary a dry run prints
- Tests: format roundtrip, witness order assembly (constructed so map iteration order disagrees with signers order), stale and tampered blob rejection, sign auto-detection, and binary tests for the argument parsing
- README: multi-signature transactions section under CLI usage

## Checklist

- [x] Builds cleanly
- [x] Tests pass (95: 89 lib, 5 binary arg tests, 1 doc test)

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,6 +349,10 @@ spel --idl program-idl.json -p program.bin \
   --co-signer <account-id> --export handover.json -- \
   admin-transfer --caller <current-admin> --evidence <new-admin> ...
 
+# --export and --co-signer are global flags and belong before the "--"
+# separator. Misplaced after it they are rejected with an error, the CLI
+# never falls through to a live submission.
+
 # The file travels to the co-signer over any channel.
 
 # Machine B: inspect and sign. Shows the summary embedded by machine A
--- a/README.md
+++ b/README.md
@@ -332,6 +332,49 @@ spel inspect 0000...0000 \
   --data 00<32-byte-definition-id-hex>00000000000000000000000000000064
 ```
 
+### Multi-Signature Transactions (Witness Exchange)
+
+Some instructions need signatures from accounts whose keys live on
+different machines. A transaction's authorization is checked against its
+witness set, so all signers must sign the same message bytes, and those
+bytes contain every signer's nonce. The CLI coordinates this through a
+partial-transaction file that travels between signers.
+
+```bash
+# Machine A: build the full transaction, sign with local keys, and write
+# the partial transaction to a file instead of submitting. Repeat
+# --co-signer for each additional required signer. Needs a node
+# connection to fetch nonces for all signers.
+spel --idl program-idl.json -p program.bin \
+  --co-signer <account-id> --export handover.json -- \
+  admin-transfer --caller <current-admin> --evidence <new-admin> ...
+
+# The file travels to the co-signer over any channel.
+
+# Machine B: inspect and sign. Shows the summary embedded by machine A
+# next to fields decoded locally from the message bytes, asks for
+# confirmation, then signs for every required signer whose key this
+# wallet holds. Works offline.
+spel sign handover.json
+
+# Anyone: submit once all witnesses are collected. Needs a node.
+spel submit handover.json
+```
+
+The file is JSON: the borsh-serialized message in hex, the ordered list
+of required signers, and the collected witnesses. The signers order
+matches the nonce order inside the message. The full format is
+documented in `spel-cli/src/blob.rs`.
+
+A partial transaction goes stale when any listed signer's on-chain
+nonce changes. Signing and submitting verify the collected witnesses
+first and fail with a clear error in that case, and the flow restarts
+with a fresh export.
+
+`--export` works without `--co-signer` too, for preparing a transaction
+on one machine and submitting it from another. It only supports public
+transactions and cannot be combined with `--dry-run`.
+
 ## Crates
 
 | Crate | Description |
--- a/README.md
+++ b/README.md
@@ -366,10 +366,20 @@ of required signers, and the collected witnesses. The signers order
 matches the nonce order inside the message. The full format is
 documented in `spel-cli/src/blob.rs`.
 
+A signature authorizes exactly the message bytes in the file being
+signed, and every already-collected witness is verified against those
+same bytes before signing, so nobody can swap the message under
+existing signatures. What the prompt cannot do is read intent into raw
+instruction data. To verify a blob independently, rebuild the claimed
+transaction on your own machine with the same arguments and `--export`,
+and compare the decoded fields, instruction data included, against
+what the prompt shows. The two exports differ only in nonces fetched
+at different times.
+
 A partial transaction goes stale when any listed signer's on-chain
-nonce changes. Signing and submitting verify the collected witnesses
-first and fail with a clear error in that case, and the flow restarts
-with a fresh export.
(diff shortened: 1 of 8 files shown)


mmlado:
Re-verified end to end on a local LEZ v0.2.0 stack (node, sequencer and wallet built from the v0.2.0 tag), this time with the program deployed. Flow: admin_transfer with a Signer candidate, partial transaction exported with --export and --co-signer, co-signed from a second wallet with spel sign, submitted with spel submit, confirmed in a block. The gated instruction then succeeds for the new admin and the sequencer rejects the old admin with Unauthorized. 
One note: the library's transfer arguments are now --new-account and --candidate after a rename in milestone 3 of the admin-authority.

danisharora099 (changes requested):
Reviewed at fe47100 as part of the RFP-001 M3 round, since the admin Signer transfer depends on this. The mechanism is sound: consistent hash-signing at this head, witness verification checks both the signature and the pubkey-to-account derivation, 104 tests green here, and the sign prompt now decodes program id, account ids and nonces from the real payload instead of trusting the summary.

One change required before I can rely on it for the transfer ceremony: **the co-signer never sees message.instruction_data**, not even as hex. That's the operative content, which instruction and which candidate for an admin transfer. As is, a malicious exporter can show a benign summary over a different payload and the co-signer cannot notice even in principle. Please print instruction_data (raw hex is fine for v1) in the "Decoded from message bytes" section; IDL-based decoding can stay the planned upgrade.

Two doc fixes: blob.rs:31-34 and README:369-372 claim signing fails on a stale blob, but sign makes no network call, staleness only surfaces at submission, please reword. And blob.rs:24-25, :58 and :108-109 still say signatures are over "exactly these bytes" while the code (correctly) signs the 32-byte message hash, which matters for third-party interop.

mmlado:
The prompt now prints the instruction data as raw hex in the decoded section. Rendered from message_hex with the same word-to-bytes encoding the dry-run captures use. An integration test now runs the real binary against a fully signed blob and pins the decode section, instruction data included.

Documentation fixes are in. Signing no longer claims to catch staleness. The signature docs now say 32-byte message hash instead of raw bytes, at the three sites you listed plus a fourth on the message_hex field itself.

mmlado:
One more hardening f93923b6be1c279bc13cf7f6182f5533f00fb8e6. Found while re-verifying the flow live.

`--export` and `--co-signer` are global flags, but nothing rejected them when placed after the -- separator. A misplaced `--export` built, signed and submitted the transaction instead of writing the handover file. It's bad flow for a command that is for off-chain signing.

The parser now validates every key after the separator against the instruction's own vocabulary. It's rejected before any transaction is built. Unit tests pin the vocabulary rules, a binary level test pins that the misplaced form refuses and writes nothing, and the README example notes the placement.

danisharora099 (approved):
Re-reviewed at f93923b. All three asks landed: the sign prompt now prints the instruction data hex (with a test), the stale-blob claims are corrected to "surfaces at submission", and the blob comments now say hash-of-bytes, with the signing itself moved to the v0.2.0 message hash. The rebase onto v0.6.0 and the unknown-flag rejection both look sound; 110 tests green on my side. Changes-requested cleared: from the RFP side this is now fit for the admin transfer ceremony. Merge timing stays with the maintainers.

vpavlin:
Rebased onto current `main` and resolved conflicts on @mmlado's behalf (head `d55a6e5`):

- **Declared conflict**: one hunk in `tests/e2e/fixture_program/Cargo.lock` — the LEZ pin (main's `v0.2.4` vs this branch's `v0.2.0`). Took main's side.
- **Hidden semantic conflict**: `exchange.rs` merged cleanly but did not compile. It landed after #256's v0.2.4 migration was written, so it was still on the pre-v0.2.4 WalletCore API. Applied the same three changes #256 made in `tx.rs` — `from_env().await`, `helm_owned()`, `poller_helm()` — plus making `sign_command` async (its only caller in `run()` already is). Committed separately as `d55a6e5`; your five commits keep their authorship.

Verified before pushing: 292 tests pass, zero build errors, `cargo fmt --check` and clippy clean. CI is now green on all nine checks — the first time this PR has had CI run at all (the runs had been sitting unapproved since Aug 9).

Test coverage is strong on the security-critical parts (`blob.rs`: tamper, wrong-key, and signer-ordering are all covered), but there is no end-to-end coverage of the export → sign → submit flow, and `submit_command` — where the v0.2.4 migration landed — is compile-checked only. Filed as #263 rather than blocking this.

One ask for #263: your manual verification against a local LEZ v0.2.0 stack predates both the v0.2.4 migration and this rebase, so a re-run of that flow on current `main` would be valuable.

danisharora099:
Confirming from the RFP-001 review side: the rebase and the v0.2.4 API adjustments in d55a6e5 do not touch the parts my approval covered (hash signing, witness verification, the decoded prompt including instruction data, the post-separator flag rejection), so the approval stands at the merged head. The re-run of the manual export, sign, submit flow on current main that you asked for is the right ask for https://github.com/logos-co/spel/issues/263; I will do that pass as part of accepting RFP-001 M3 rather than leaving it only to the grantee.

PR #253, opened 2026-08-01
fix(idl-gen): skip path-dep source files the consumer edition cannot re-lex

mmlado:
## Description

`generate_idl!` walks path dependency sources with `syn::parse_file`. Inside a proc macro this routes through rustc's lexer. When a walked file contains a macro metavariable glued to a string literal, like `stringify!($Fp"({})")`, the lexer queues a fatal `prefix is unknown` error as a side effect. The walk catches the parse error and continues, but the build is already failed, with an error that points at the macro invocation and names a token that appears nowhere in the consumer's code. The shape is legal in editions before 2021. ark-ff 0.3.0 ships exactly this line (src/fields/macros.rs, line 615), so a pre-2021 path dependency can break a consumer that did nothing wrong.

Repro on current main: a program file, a `generate_idl!("program.rs")` bin, and an edition 2018 path dependency containing `write!(f, stringify!($Fp"({})"), self.0)` inside a `macro_rules!`. `cargo check` fails with `error: prefix 'Fp' is unknown`
pointing at the `generate_idl!` call. Changing that one line to `write!(f, "{}({})", stringify!($Fp), self.0)` makes the same project compile, so the metavariable to literal adjacency is the exact trigger.

## Changes

- `spel-framework-core/src/idl_gen.rs`: skip source files containing a macro metavariable glued to a string literal before parsing them. Such files cannot carry usable `#[account_type]` items.
- Two regression tests. `landmine_file_contributes_no_items_but_is_tracked` fails without the gate and passes with it.

## Checklist

- [x] Builds cleanly
- [x] Tests pass (`cargo test -p spel-framework-core --features idl-gen`, 112 passed)
- [ ] README updated if new features, CLI commands, or behaviour changed (bug fix only, no surface change)
- [x] New public methods have doc comments (the new helper is private and documented)

### DIFF
--- a/spel-cli/src/generate_idl.rs
+++ b/spel-cli/src/generate_idl.rs
@@ -170,7 +170,7 @@ mod tests {
 
     impl Drop for TempDir {
         fn drop(&mut self) {
-            let _ = fs::remove_dir_all(&self.0);
+            fs::remove_dir_all(&self.0).ok();
         }
     }
 
@@ -823,7 +823,8 @@ pub mod token {
         );
         assert_eq!(dep_result.dirs.len(), 1);
 
-        let idl = generate_idl_from_file_with_deps(&program, &dep_result.dirs).unwrap();
+        let idl =
+            generate_idl_from_file_with_deps(&program, &dep_result.dirs, &mut |_| {}).unwrap();
 
         assert_eq!(idl.name, "token");
         assert_eq!(idl.instructions.len(), 4);
@@ -906,7 +907,8 @@ pub mod token {
             "unexpected warnings: {:?}",
             dep_result.warnings
         );
-        let idl = generate_idl_from_file_with_deps(&program, &dep_result.dirs).unwrap();
+        let idl =
+            generate_idl_from_file_with_deps(&program, &dep_result.dirs, &mut |_| {}).unwrap();
 
         let account_names: Vec<&str> = idl.accounts.iter().map(|a| a.name.as_str()).collect();
         assert!(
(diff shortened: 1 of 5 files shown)


danisharora099 (commented):
Reviewed at be37ef2 alongside the RFP milestone round. The fix is correct and the trigger is real, I confirmed the ark-ff 0.3.0 landmine on disk; 112 tests green locally with the idl-gen feature. One ask: the skip is currently silent, and the byte-scan is context-free, so a `$word"` inside an ordinary string or doc comment in a dep would drop that whole file from account-type harvesting with no signal. Please route the skip through the on_warning channel (or expose a files_skipped output) so it's at least visible. With that, good to land from my side; maintainers' call on the rest.

mmlado:
Both suggestions applied. The on_warning callback is now threaded through like in find_path_dep_dirs, the CLI prints it to stderr and the macro path prints it during expansion. 

It is now context-aware: the scan now tracks strings, raw strings, comments and char literals, so a $word" inside a doc comment or string no longer drops the file. Anything the scan cannot classify still counts as code.

The scanner moved to idl_gen/unlexable.rs with its own tests.

danisharora099 (commented):
The visibility ask landed nicely at b260a51: skips now route through on_warning, the heuristic is context-aware (string and comment matches ignored), and rebuild tracking is preserved. Good to land from my side.

vpavlin:
Reviewed — this fixes the reported ark-ff trigger and strictly improves on main, so not blocking. During review we empirically confirmed some detection gaps in `unlexable.rs` (missed reserved-prefix shapes that still abort the build, and two parity-desync cases incl. `cr#"..."#` literals); documented with repros in #258 as agreed follow-up.

PR #257, opened 2026-08-19
 feat(framework): the SPEL extension mechanism

mmlado:
## Description

Merges the SPEL extension mechanism into main in one step. The work was developed and reviewed on `feat/admin_authority_m3`, proven by three extension libraries built on it, and the branch carries the v0.6.0 release merge so main's history is fully contained.

The framework knows nothing about any specific extension. An extension is a normal crate that declares itself in Cargo metadata, and the framework wires it into a consumer program at compile time.

The mechanism ships proven by three extension stacks, merging to their mains in lockstep:

- spel-authority, the shared slot primitive: https://github.com/mmlado/spel-authority/pull/5
- spel-admin-authority, single-admin authority with gates: https://github.com/mmlado/spel-admin-authority/pull/7
- spel-freeze-authority, program and per-account freezing, composing with admin in one shared account: https://github.com/mmlado/spel-freeze-authority/pull/6

The mechanism was reviewed piece by piece in the PRs this replaces, closing with this one:
- https://github.com/mmlado/spel/pull/1 gate param injection
- https://github.com/mmlado/spel/pull/2 IDL arg types, CLI enum args
- https://github.com/mmlado/spel/pull/3 the wrapper kwarg contract and IDL parity
- https://github.com/mmlado/spel/pull/4 embedded account support and cross-marker bound args
- https://github.com/mmlado/spel/pull/5 fork CI proof run for --spel-git
- https://github.com/mmlado/spel/pull/6 idl-gen visibility, the unmatched-marker hard error, range-overlap windows, creator auto-exemption

## Changes

- **Discovery.** A marker attribute on the `#[lez_program]` module matches a direct dependency declaring `extension_attr` in its `[package.metadata.spel]`. Path, git and registry dependencies are discoverable, transitive ones never are. A marker matching nothing is a hard compile error whatever the reason it did not match.
- **Instruction merging.** A discovered extension's `#[instruction]` fns join the consumer's dispatcher, `Instruction` enum, and IDL, dispatched by direct cross-crate call. Marker order is the cross-extension ABI.
- **Gate injection.** Per-instruction gate attributes get their account params synthesized from inject metadata unless the consumer already declares them, with declared params recognized by role. The wrapper kwarg contract keeps the extension macro and the framework in agreement, enforced by an alignment self-test pattern.
- **Auto-wrap.** An extension can gate every dispatched instruction, with a manual opt-out word, a per-fn exempt marker, a cross-extension exempt list, and an automatic carve-out for the instruction that creates a gate's embedded account.
- **Embedded mode.** An extension's fixed-size config can live inside a consumer account at a declared byte offset instead of a dedicated PDA. Offsets compile into the program and never appear in the IDL or a transaction, window collisions between embedded extensions refuse to compile through the mandatory `state_type` size, two extensions can share one consumer account with the shared account merged to a single transaction entry, and cross-marker bound args let one extension read state a peer embedded without depending on the peer's crate.
- **IDL and CLI.** `generate-idl` resolves the consumer's dependency graph for account type collection, resolves type aliases, and shows injected accounts and gates. `spel init` scaffolds a consumer and follows the repository context for its framework URL.

## Checklist

- [x] Builds cleanly (`cargo build --workspace`)
- [x] Tests pass (`cargo test --workspace`, e2e included)


vpavlin:
Reviewed this by merging it onto current `main`, then building a real extension from the README alone and running it against a live LEZ v0.2.4 sequencer. Branch with everything: **[`review/ext-mechanism`](https://github.com/logos-co/spel/tree/review/ext-mechanism)**.

## It works, end to end

I wrote a pause-switch extension — marker attr, gate attr, two contributed instructions, one injected account — and drove it on chain:

| step | result |
|---|---|
| `initialize` (consumer's own) | confirmed |
| `init_pause` — extension instruction | confirmed |
| `do_something` while unpaused | confirmed |
| `set_paused true` — extension instruction | confirmed |
| `do_something` while **paused** | **rejected**: `Program error 1001: program is paused` |
| `set_paused false` | confirmed |
| `do_something` again | confirmed |

Discovery, instruction merging into the consumer's dispatcher, account injection into the gated instruction, and gate enforcement all work. Steps 3 → 5 → 7 show the gate is genuinely state-dependent, not sticky. Both IDL producers (`generate_idl!` and `spel generate-idl`) emit identical instruction lists and identical injected-account ordering.

Other things worth calling out as good:

- **The trust model doc is excellent** — names the threat, states the decision, justifies it against cargo's own boundary, records rejected alternatives. It also made the review tractable, because it states falsifiable claims I could check against the code.
- **The marker-without-dependency error is the standard other failure paths should meet.** Removing the dep while leaving the marker gives: `marker(s) ["pause_ext"] matched no discoverable extension and dependency resolution failed: … refusing to compile a program that could be silently missing its extension surface.` It names the marker, explains the cause, and says why it refuses.

## Merge, don't rebase

Merging `main` in resolved with **zero conflicts**, and the result builds clean with **434 tests passing**. I tried a rebase first and abandoned it — 13 commits each colliding with #253's `unlexable`/`on_warning` changes, six manual resolutions in. The merge commit is on the branch above if it's useful.

## DX suggestions

Full write-up in [`docs/extension-dx-feedback.md`](https://github.com/logos-co/spel/blob/review/ext-mechanism/docs/extension-dx-feedback.md) on that branch. The short version:

**1. `generate_idl!` discards discovery warnings (one line).** `spel-framework-macros/src/lib.rs:235` and `:2360` pass `&mut |_| {}`, while the CLI passes `|w| eprintln!("⚠️  {w}")`. The trust model promises environmental problems are "reported through the discovery `on_warning` channel" — in the macro path they're swallowed, so during a normal `make idl` the author gets silence.

**2. Three authoring requirements are load-bearing but undocumented.** Each stops an author cold, and I only got past them by reading `spel-admin-authority`:
- you must ship your own `#[instruction]` proc-macro that strips `#[account(...)]` param attrs, or the extension crate won't compile — ~8 lines of boilerplate every author reinvents (could the framework export it to re-export?)
- `extern crate self as <crate>;` is required for contract 2's absolute self-paths to resolve
- `SpelOutput::execute` needs explicit `(Account, AutoClaim)` tuples, since `ExecuteTransformer` only runs inside `#[lez_program]`

**3. No in-repo example of an instruction-contributing extension.** `mini_ext` deliberately has none (it exists for the compile-failure fixtures), so the only working reference is an external repo. I put my pause extension on the branch at [`docs/examples/pause-extension/`](https://github.com/logos-co/spel/tree/review/ext-mechanism/docs/examples/pause-extension) — take it, adapt it, or ignore it. It carries the three requirements above by demonstration.

**4. Path-dep extensions must live inside the project tree** — the guest build is docker-hermetic, so a path outside it fails with `failed to load manifest`. Worth a sentence.

**5. Adjacent, not yours:** `spel init` emits independent framework pins for `methods/guest` and `examples`. If they drift to different versions, `make idl` produces an IDL missing the entire extension surface, silently — and that IDL is what gets committed and fed to client-gen. Pre-existing scaffold behaviour, but the extension mechanism is what makes the consequence severe. Worth pinning them in lockstep.

## Not your bugs

Two failures I hit while writing the extension were mine, not the framework's, and I'm noting them only because other authors will hit them too: `#[account(init, ...)]` on a setter → `AccountAlreadyInitialized` on the second call, and claiming an already-initialised account → `ClaimedNonDefaultAccount`. Both were caught by LEZ with errors that named the exact rule. They're evidence for suggestion 2's third bullet — hand-writing claim semantics outside `#[lez_program]` is easy to get wrong.

Still untried: the malformed-metadata and duplicate-instruction-name paths. Happy to run those if useful.

vpavlin:
Small correction to suggestion 5 above, in your favour: I checked `init.rs` and `spel init` emits the **same** `{spel_ref}` for both `methods/guest` (:719) and `examples` (:810, :812), so they are consistent at scaffold time — I implied otherwise and that was wrong.

The real risk is narrower: they are separate crates with separate lockfiles on a *floating* branch ref, so they can resolve to different commits later. That is what I hit. Filed separately as #269 since it is independent of this PR.

Also filed #270 for adopting a positive extension example, blocked on this landing — with the suggestion that it live under `tests/e2e/` so CI compiles it rather than sitting in `docs/` where it would rot. The candidate on the review branch is yours to take, adapt, or replace.

mmlado:
All three of these are documented, in the extension author guide in logos-co/logos-docs#355, which is open and has been through a long dogfooding loop of its own. The #[instruction] shim is given verbatim with the reason it exists, extern crate self as is in the runtime-library snippet with an explanation of why the path has to resolve in both compiles, and the post-state rule is there including the part that bit you, that the vec is homogeneous, so any claimed account turns every entry into a tuple with AutoClaim::None for the rest.

So the gap is real but it is one of placement rather than of writing: the material exists and is not merged, and it lives in the docs site rather than in this repo. Once it lands, the README probably wants to link it rather than restate it.

Your parenthetical is a separate question and documentation does not answer it. Every author still copies those eight lines by hand, and exporting the shim from the framework for extensions to re-export would remove the boilerplate rather than explain it. Worth filing on its own.

vpavlin:
Heads up — I pushed two commits to this branch (maintainer edits are on). Both are integration housekeeping from our merge order, not review changes; nothing of yours is modified and nothing was force-pushed, so a plain `git pull` will take them.

**`cb7e780` — merge `main`.** Conflict-free, as before. Brings in #267, #268, #272 and #266.

**`5011446` — sixteen test call sites.** This is the part worth knowing about. #266 (Vec<u64>/Vec<u128>/Vec<bool> CLI parsing) landed this morning with tests calling `parse_value(raw, &ty)`. This branch adds a third `types: &[IdlTypeDef]` parameter to that function. The two merge with **no textual conflict** and the test build then fails with sixteen

```
error[E0061]: this function takes 3 arguments but 2 arguments were supplied
```

Both sides were correct in isolation — each was tested against a `main` that did not yet contain the other. I caught it by trial-merging the two before either landed, rather than after.

`&[]` is right at every site: they exercise `Vec<primitive>` arguments, which resolve without consulting the IDL's type definitions. No assertion changed, so #266's tests still check what they were written to check, and no production code is touched. `cargo fmt` reflowed one call that grew past the line limit.

Verified on the merged tree before pushing: `cargo test -p spel` → 112 passed, `cargo fmt --check` clean, no new clippy warnings.

**Why we did it rather than asking you:** #266 comes from an org-owned fork, where GitHub does not offer maintainer edits at all — so that branch could only ever be fixed by its author, while this one we could just do. Merging #266 first and absorbing the fix here was the only order that did not cost a third round-trip to a contributor who had already done two.

**One thing that is yours to decide:** `parse_value` is a public signature change, and #266 will not be the last PR in flight that calls it. If you kept a two-argument wrapper delegating to the three-argument version, in-flight branches would keep compiling and pick up the new behaviour when they rebase. Entirely your call — I did not want to make an API decision inside your PR.

Sorry for the noise on your branch; shout if you would rather I had left it and let CI tell you.

mmlado:
Pulled and checked both commits. The sixteen call sites are right, every one is a Vec of a primitive, and those never consult the type defs, so &[] matches what production passes there. Ran the full workspace suite on 5011446 locally, everything green, fmt clean, and the e2e jobs on the PR have all finished green too.

On the wrapper, I would keep the break loud. A two argument wrapper passing &[] keeps in-flight branches compiling, but it also lets a new call site that actually deals in custom types compile and then fail at parse time instead of at build time. The E0061 error costs one rebase and is gone, the wrapper would stay in the API. Anyone who hits it has the same one line fix you just applied sixteen times.

Good to merge from my side. Thank you for fixing the #266 collision

PR #276, opened 2026-09-15
fix(macros): the standalone #[instruction] strips #[account] attrs

### LINKED ISSUE #271: Extension authors must hand-write an `#[instruction]` shim because the framework's own is a bare pass-through
Every SPEL extension library ships its own ~8-line `#[instruction]` proc macro whose only job is to strip `#[account(...)]` attributes off handler parameters. The framework already exports an `#[instruction]`, and already contains the stripping logic — but the two are not connected, so each author reimplements it identically.

Raised by @mmlado on #257: *"Every author still copies those eight lines by hand, and exporting the shim from the framework for extensions to re-export would remove the boilerplate rather than explain it. Worth filing on its own."*

## Why the obvious thing doesn't work

`spel_framework_macros::instruction` ([lib.rs:111](https://github.com/logos-co/spel/blob/main/spel-framework-macros/src/lib.rs#L111)) is a bare pass-through:

```rust
#[proc_macro_attribute]
pub fn instruction(_attr: TokenStream, item: TokenStream) -> TokenStream {
    item
}
```

That is correct for its documented purpose — inside `#[lez_program]` it never expands at all, because the module-level macro consumes the whole module first and strips both `#[instruction]` and `#[account(...)]` itself in `generate_handler_fns` ([lib.rs:1225-1230](https://github.com/logos-co/spel/blob/main/spel-framework-macros/src/lib.rs#L1225-L1230)).

But an extension library's handlers live *outside* `#[lez_program]`, so there the attribute does expand, and a pass-through leaves the `#[account(...)]` attrs for rustc:

```rust
use spel_framework_macros::instruction;

#[instruction]
pub fn set_paused(
    #[account(mut, pda = literal("pause"))] state: u32,
    #[account(signer)] admin: u32,
    paused: bool,
) -> u32 { 0 }
```

```
error: cannot find attribute `account` in this scope
error: cannot find attribute `account` in this scope
error: could not compile `shim-test` (lib) due to 2 previous errors
```

So the author is pushed into standing up a `proc-macro = true` sub-crate with `syn`/`quote`/`proc-macro2` and rewriting the strip by hand. That is what `spel-admin-authority` does, what my pause extension on `review/ext-mechanism` does, and what the extension-author guide (logos-co/logos-docs#355) documents verbatim.

## Proposed fix

Have the exported `#[instruction]` do the strip when it expands:

```rust
#[proc_macro_attribute]
pub fn instruction(_attr: TokenStream, item: TokenStream) -> TokenStream {
    let mut func = parse_macro_input!(item as ItemFn);
    for input in &mut func.sig.inputs {
        if let FnArg::Typed(pat_type) = input {
            pat_type.attrs.retain(|a| !a.path().is_ident("account"));
        }
    }
    quote!(#func).into()
}
```

Verified locally:

- the failing example above compiles with it,
- `cargo test -p spel-framework-macros -p spel-framework-core` stays green (130 tests across both).

It is inert for the `#[lez_program]` path, which never expands this macro, so the change only affects standalone use — which today is broken anyway.

Extension libraries could then `pub use spel_framework::instruction;` alongside their marker attribute, and drop the sub-crate entirely unless they also ship a gate attribute.

## Scope

Independent of #257 — applies whether or not the extension mechanism lands, since standalone `#[instruction]` is documented as supported (`docs/reference/macros.md`: *"a no-op when used standalone"*) and is currently unusable with `#[account(...)]` params.

Worth deciding alongside: whether `docs/reference/macros.md` should say the standalone behaviour strips helper attributes, rather than calling it a no-op.

mmlado:
## Description

`spel_framework_macros::instruction` was a bare pass-through. That is right inside `#[lez_program]`, where it never expands at all, because the module macro consumes the whole module first and emits the handlers itself. An extension library's instruction fns live outside `#[lez_program]`, so there the attribute does expand, and a pass-through leaves the `#[account(...)]` parameter attrs for rustc:

```
error: cannot find attribute `account` in this scope
```

The way out so far has been a `proc-macro = true` sub-crate with syn, quote and proc-macro2 that strips the attrs by hand. `spel-admin-authority` ships one, and the extension-author guide documents it.

Closes #271.

## Changes

- `#[instruction]` now parses the function and drops the `#[account(...)]` attrs off its parameters when it expands.
- The strip moves into `strip_account_attrs`. `generate_handler_fns` already carried its own copy of those four lines, so both emitters now share one.
- New integration test `spel-framework-macros/tests/standalone_instruction.rs`, which is the issue's failing example. The file compiling is the regression guard. Four unit tests cover the helper directly: it clears every `account` attr, keeps attrs it does not own, leaves function-level attrs alone, and accepts a parameterless fn.
- README gains contract 4 for extension authors: re-export `pub use spel_framework::instruction;` instead of hand-writing a stripping shim.

The change is inert for the `#[lez_program]` path, which never expands this macro. Discovery is unaffected too, because the extension scanner parses the library's source file rather than this expansion, so it still sees the attrs.

## Verification

- `cargo test -p spel-framework-macros -p spel-framework-core`: 267 pass, 0 fail.
- `cargo fmt --all --check` clean. Clippy reports no new warning.
- Cross-repo check: `spel-admin-authority` with its own `instruction` shim deleted and `pub use spel_framework::instruction;` in its place checks clean, and its full suite passes at 53 tests, including all three IDL golden tests and the `compile_fail` case. Run against the framework revision that repo pins rather than `main`, since the two lineages differ.

## Open point from the issue

The issue asks whether `docs/reference/macros.md` should say the standalone behaviour strips helper attributes rather than calling it a no-op. That file does not exist in this repo or in logos-docs, so the macro's own doc comment and the README carry the wording instead. Happy to follow up in logos-co/logos-docs#355 if the extension-author guide should drop its shim recipe.

## Checklist
- [x] Builds cleanly
- [x] Tests pass


vpavlin:
Tested this end to end against the pause extension from `review/ext-mechanism`, with its hand-written `instruction` shim deleted and `pub use spel_framework::instruction;` in its place:

| check | result |
|---|---|
| extension built against `main` | fails — `cannot find attribute 'account' in this scope` (negative control) |
| extension built against `main` + this PR | builds |
| `spel generate-idl` on a consumer | all 4 instructions discovered, `pause_config` injected with its seeds, no warnings |
| guest build (docker, pinned to `8dfb80e`) | builds |
| on a local LEZ v0.2.4 sequencer | `init_pause` ✓, `bump` ✓, `set_paused true` → `bump` rejected with `1001: program is paused` ✓, `set_paused false` → `bump` ✓ |

Also checked `main` + this PR together (#274 touched the same file after CI ran here): macros/core/framework tests pass, fmt and CI-style clippy clean.

LGTM 👍

mmlado:
Thank you for the merge. This will make writing libraries simpler.

